PR #13, opened 2019-08-23
Adding mining functionality for epoch and state

agatsoh:
Allows us to mine to next Epoch or next state in the tests so no need to manually change contracts just for setting epoch and state.


agatsoh:
tests only passes for `StakeManager.js` if this gets approved then would work on the other tests.

agatsoh:
Implemented for all tests @hrishikeshio  please have a look.

Issue #35, opened 2020-08-04
move build/ out of repo


hrishikeshio:
build can be its own repo. All other razor network repos can refer to that folder.

this needs all dependent razor network repos to be modified.

SkandaBhat:
- Import build folder from its own repos (Contracts and Bridge).
- Refactor throughout projects
- Delete/add to .gitignore build folders throughout projects

SkandaBhat:
We have npm module now, this issue is fixed.

PR #68, opened 2021-01-25
Build as Sub Repo

abhishekvispute:
As Build is required between multiple repos, like cli, bridge.
It makes sense in making it as subrepo
As pointed out in https://github.com/razor-network/contracts/issues/35

Now to do that there are various options
1) Using Submodules or Subtress: Not used given added complexity for a end-user
2) Including push to dependent repos via CI : Not used as it increases CI exec time
3) Having build under .gitignore and manually updating it in respective dependent repos : It's true that build is something that does not change repeatedly, so having it in .gitignore makes sense there. But as some contracts are now in the development stage, this is not a good choice
4)  Using [git-subrepo](https://github.com/ingydotnet/git-subrepo) 🥇  This is the best option as it gets works done without any additional efforts needed for end-user

For contributors, its as simple as following  👍 
   - The upstream : [razor-network/build](https://github.com/razor-network/build)
   - [Installation Instrusctions](https://github.com/ingydotnet/git-subrepo#installation)
   - To pull from upstream : (from parent repo)  git subrepo pull build
   - To push to upstream : (from parent repo) git subrepo push build

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,3 +44,15 @@ These are the contracts for Razor network.
 # Addresses
 Contract addresses can be found [here](ADDRESSES.md)
 
+# Guidelines for Collaborators
+
+- Build 
+   - We use [git-subrepo](https://github.com/ingydotnet/git-subrepo) for Build, as its common between multiple repos.
+   - The upstream can be found at [razor-network/build](https://github.com/razor-network/build)
+   - [Installation Instrusctions for git-subrepo](https://github.com/ingydotnet/git-subrepo#installation)
+   - To pull from upstream : (from parent repo)  git subrepo pull build
+   - To push to upstream : (from parent repo) git subrepo push build
+
+
+
+   
--- a/build/.gitrepo
+++ b/build/.gitrepo
@@ -1,0 +1,11 @@
+; DO NOT EDIT (unless you know what you are doing)
+;
+; This subdirectory is a git "subrepo", and this file is maintained by the
+; git-subrepo command. See https://github.com/git-commands/git-subrepo#readme
+;
+[subrepo]
+	remote = none
+	branch = master
+	commit = 
+	method = merge
+	cmdver = 0.4.3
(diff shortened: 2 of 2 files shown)


dev1644:
This PR looks good to me, but we should rename `build` to something more specific like `ABIs` or `contracts_artificats`, as `build` is more generic and doesn't give the idea what's it's about. Let me know about your thoughts too.

SkandaBhat:
> This PR looks good to me, but we should rename `build` to something more specific like `ABIs` or `contracts_artificats`, as `build` is more generic and doesn't give the idea what's it's about. Let me know about your thoughts too.

Sounds good, we can do this.

dev1644:
Since changes in `Contract Addresses` will be there along with ABIs, we should add contract addresses too.

abhishekvispute:
> This PR looks good to me, but we should rename `build` to something more specific like `ABIs` or `contracts_artificats`, as `build` is more generic and doesn't give the idea what's it's about. Let me know about your thoughts too.

Hi @dev1644 ,
The thing is currently we use truffle, and it expected artifacts in the Build folder.
Now suppose if we rename this to ABI, I am not sure if we can configure truffle to recognize these files.
Is there a way?

Okay and your contract address topic, I am not clear, what you are mentioning
Does it mean that we should mention that we also have deployed contract addresses in this repo in description or readme?

Thanks

dev1644:
Hi @abhishekvispute 
>The thing is currently we use truffle, and it expects artifacts in the Build folder.

If I am doing local dev work using Truffle, isn't `build` gonna change for every compilation/test? If yes then we need to move/rename the `ABIs` into a different folder (This folder will be part of subrepo). 

> Okay and on your contract address topic, I am not clear, what you are mentioning
Does it mean that we should mention that we also have deployed contract addresses in this repo in description or readme?

ABIs are subject to change either when there are changes into contracts or we update Solc compiler setting (like updating compiler version or optimizer etc). So assuming we are not gonna change compiler setting often, changes in smart-contracts will mean that new addresses and other apps will need both ABIs and addresses to interact with contracts.

SkandaBhat:
Benching this PR for now. We will bring this up at a later stage. Stale PRs accumulate conflicts.

Issue #9, opened 2019-08-15
blockManager.confirmBlock() called each time staker commits


agatsoh:
We can create a switch inside the VoteManager contract which gets switched so that we dont need to call this function each time the staker commits.

SkandaBhat:
This is merged in #69. Closing this.

PR #69, opened 2021-01-30
Optimization : Switch for Confirm Block in commit

abhishekvispute:
As Required in Issue https://github.com/razor-network/contracts/issues/9 

1. Added getNumProposedBlocks in IBlockManager Interface.
2. Added Switch in Vote Manager confirm block only when block in previous epoch has not been confirmed and if previous epoch do have proposed blocks
3. Removed Redundant switch from BlockManager due to 2

Thanks 👍


SkandaBhat:
Looks good to me. @SamAg19 Please look at this.

SamAg19 (approved):

Issue #66, opened 2021-01-14
To add decreasing BlockReward functionality


abhishekvispute:
As of now 
The block reward is constant 
which should slowly decrease over time

```
function blockReward() public pure returns(uint256) { return(40*(10**uint256(18)));} 
```

abhishekvispute:
Closed by https://github.com/razor-network/contracts/pull/72

PR #72, opened 2021-02-10
To Implement Block Reward Function | Halving

abhishekvispute:
Hi All,
As pointed out in https://github.com/razor-network/contracts/issues/66 we needed a decreasing block reward function.
The same has been implemented in this PR.

We have used the standard halving function.
Where we halve block reward after a specific interval (measured in no of blocks)
```
  function _calculateBlockReward() private returns(uint256)
    {
        uint256 halvings = (block.number - genesisBlock) / Constants.halvingInterval();	
        if(halvings > lastHalvings) {
             lastBlockRewards = Constants.initialBlockReward() >>  halvings;
             lastHalvings = halvings;
        }
	    return lastBlockRewards	;		
    }
   
   ```
   **Please note that we will have to tune these constants (halvingInterval and Initial Block Rewards) as per tokenomics and deployment chain.**
   
   As of now, I have kept _calculateBlockReward() as private but I don't think visibility makes any difference here. 
   So kept Private just to be extra secure :)
   
  The new test has been added to test this functionality "BlockReward.js"
  It's skipped by default as it takes a very long time to pass with current defaults.
  _It takes long time as we have to mine blocks till we reach block reward 0, to check overflow conditions_
  (All details are mentioned in the source file)
  
  Please let me know if this is right approach 
  
  Thanks 👍

### DIFF
--- a/contracts/Core/StakeManager.sol
+++ b/contracts/Core/StakeManager.sol
@@ -166,6 +166,7 @@ contract StakeManager is Utils, ACL, StakeStorage {
         uint256 halvings = (block.number - genesisBlock) / Constants.halvingInterval();	
         if(halvings > lastHalvings) {
              lastBlockRewards = Constants.initialBlockReward() >>  halvings;
+             lastHalvings = halvings;
         }
 	    return lastBlockRewards	;		
     }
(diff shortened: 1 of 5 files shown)


SkandaBhat:
Looks good to me. @hrishikeshio @SamAg19 Please have a look.

SamAg19 (approved):

PR #87, opened 2021-03-24
Fix issue #84

### LINKED ISSUE #84: Issue to change variable name in StakeManager.sol contract & to check the test case in test file "StakeManager.js" for the same
**_Issue 1)_**  Suppose there are 4 stakers  if a staker  (who is not 4th staker) is staking again for second time his unstake after & withdrawAfter must be updated but here **_in line 15 & 16_**  the unstake after & withdrawAfter of 4th staker will be updated because numstakers=4. 
**_Issue 2)_** And also  in file "stakemanager.js"  the test case "should be able to unstake after unstake lock period" is passing because in test case "Should be able to increase stake"  we are making the 1st staker to stake again (2nd time) in the same epoch so the unstake after & withdrawAfter of the 1st staker won't get updated but still the unstake after & withdrawAfter for 1st staker will  be (currentEpoch+1) so in next epoch he can unstake. But if we make 1st staker to stake in next epoch the test case "should be able to unstake after unstake lock period" fails. 
 
                function stake (uint256 epoch, uint256 amount) external checkEpoch(epoch) checkState(Constants.commit()) {
                require(stateManager.getState() != Constants.reveal(), "Incorrect state");
                require(amount >= Constants.minStake(), "staked amount is less than minimum stake required");
                require(sch.transferFrom(msg.sender, address(this), amount), "sch transfer failed");
                uint256 stakerId = stakerIds[msg.sender];
                uint256 previousStake = stakers[stakerId].stake;
                if (stakerId == 0) {
                 numStakers = numStakers.add(1);
                stakers[numStakers] = Structs.Staker(numStakers, msg.sender, amount, epoch, 0, 0,
                epoch.add(Constants.unstakeLockPeriod()), 0);
                stakerId = numStakers;
                stakerIds[msg.sender] = stakerId;
                 }  else {
                          stakers[stakerId].stake = stakers[stakerId].stake.add(amount);
**_15)_**            stakers[numStakers].unstakeAfter = epoch.add(Constants.unstakeLockPeriod());
**_16)_**            stakers[numStakers].withdrawAfter = 0;
                 }
                 emit Staked(epoch, stakerId, previousStake, stakers[stakerId].stake, now);
                 }

GauravJain9:
This is a PR  for issue #84 
Fixes #84
Added a separate test case to check that unstakeAfter and WtihdrawAfter are working properly as mentioned in issue #84      

Task Done:
1) Changed the variable **_numstakers to stakerId_** in **_StakeManager.sol file_** because of issue #84 
2) Added test case **_"Should be able to reset the lock periods"_**  of **_StakeManager.js file_** because of issue #84 
3) Added a line **_"mineToNextEpoch"_** in **_"should be able to increase stake"_** test case of **_StakeManager.js file_** because of issue #84 

**_1st change is required since_** it was updating both unstakeAfter and withdrawAfter for only the last staker Eg. If we have 5 stakers then both unstakeAfter and withdrawAfter were being updated for the 5th staker only no matter whoever stakes again & again.

**_2nd  task "test case addition" is required_** since it tests if both unstakeAfter and withdrawAfter are reset only for the particular staker who stakes again instead of resetting it for the last staker.

**_3rd task reason:_**
Earlier the test cases were passing because we were making the 1st staker to stake again in "should be able to increase stake" in the same epoch so each time the unstakeAfter will always be the  (epoch+1) and assertions were true. 
But now if we make 1st staker to stake in the nextEpoch then assertion becomes false because the unstakeAfter and withdrawAfter of last staker is being updated and not for the 1st staker.


SamAg19 (approved):

SkandaBhat:
Please merge latest changes from master into this branch.

SkandaBhat (approved):

PR #93, opened 2021-04-06
updated contract compiler version to v0.8.0 & removed unused dependencies from package.json

adi44:
Removed Safemath.sol
Changed "now" keyword to "block.timestamp"
Changed path for oppenzeppelin/merkle.js
removed pragma ABI encoder
Changed openzeppelin-contracts to @openzeppelin/contracts_0.4.0

### DIFF
--- a/contracts/Bridge.sol
+++ b/contracts/Bridge.sol
@@ -1,4 +1,4 @@
-pragma solidity 0.6.11;
+pragma solidity ^0.8.0;
 // pragma experimental ABIEncoderV2;
 // import "../SchellingCoin.sol";
 // import "./Utils.sol";
@@ -8,7 +8,7 @@ pragma solidity 0.6.11;
 // import "./IVoteManager.sol";
 // import "./JobStorage.sol";
 // import "./IStateManager.sol";
-// import "openzeppelin-solidity/contracts/math/SafeMath.sol";
+
 // import "./Core/WriterRole.sol";
 
 
--- a/package.json
+++ b/package.json
@@ -31,9 +31,9 @@
   },
   "homepage": "https://github.com/razor-network/contracts#readme",
   "dependencies": {
+    "@openzeppelin/contracts": "^4.0.0",
     "@razor-network/merkle": "^0.6.0",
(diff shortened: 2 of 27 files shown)


contracts/Bridge.sol
```diff
@@ -8,7 +8,7 @@ pragma solidity 0.6.11;
 // import "./IVoteManager.sol";
 // import "./JobStorage.sol";
 // import "./IStateManager.sol";
-// import "openzeppelin-solidity/contracts/math/SafeMath.sol";
```
dev1644, line 11: Since we won't be needing `SafeMath` (at v0.8.0+), let's remove this import @adi44

package.json
```diff
@@ -31,6 +31,7 @@
   },
   "homepage": "https://github.com/schelling-network/contracts#readme",
   "dependencies": {
+    "@openzeppelin/contracts": "^4.0.0",
     "@razor-network/merkle": "^0.6.0",
     "jsonfile": "^6.1.0",
     "openzeppelin-solidity": "^3.1.0"
```
dev1644, line 37: Can we remove `openzeppelin-solidity` dependency since we are not using it anymore? @adi44

dev1644 (changes requested):

package.json
adi44: I have removed it from the package.json @dev1644

contracts/Bridge.sol
adi44: removed this  import line @dev1644

dev1644:
Hi @adi44 , here, we have 9 commits for this PR, changes aren't drastic that much, so it would be better if you just rebase & squash the commits into a single commit with the message `updated  contract compiler version to v0.8.0 & removed unused dependencies from package.json`.

SamAg19 (approved):

dev1644 (approved):

abhishekvispute (approved):

SamAg19 (approved):

SkandaBhat:
Fixes #73

PR #100, opened 2021-04-18
Constant Configurable Block Rewards by Admin 

abhishekvispute:
Hey Hi,
This PR 
Fixes https://github.com/razor-network/contracts/issues/74.

As per the issue, a constant block reward is needed which admin can update anytime.
We achieved so by constructor and update defined in StakeManager.
The update is restricted to admin only by ACL.
Removed calculation of block reward as per the curve.

Thanks


test/helpers/constants.js
```diff
@@ -7,6 +7,7 @@ const EPOCH_LENGTH = BigNumber.from(40);
 const NUM_BLOCKS = 10;
 const NUM_STATES = BigNumber.from(4);
 const STATE_LENGTH = BigNumber.from(10);
+const BLOCK_REWARD = BigNumber.from(40).mul((BigNumber.from(10).pow(BigNumber.from(18))));
```
dev1644, line 10: > BigNumber.from(40).mul((BigNumber.from(10).pow(BigNumber.from(18))));

Can be re-written as `BigNumber.from(40).mul(ONE_ETHER);`

test/ACL.js
```diff
@@ -245,6 +246,11 @@ describe('Access Control Test', async () => {
     await stakeManager.revokeRole(await constants.getStakerActivityUpdaterHash(), signers[0].address);
   });
 
+  it('Only Default Admin should able to update Block Reward', async () => {
+    await assertRevert(stakeManager.connect(signers[1]).updateBlockReward(100), expectedRevertMessage);
+    assert(await stakeManager.updateBlockReward(100), "Admin not able to update BlockReward");
```
dev1644, line 251: Let's add one more case, where it `updateBlockReward` will be called by `non-admin` account & it should revert. @abhishekvispute

abhishekvispute: Yes, that's what line no. 250 does @dev1644

test/helpers/constants.js
abhishekvispute: Done, Thanks

test/ACL.js
dev1644: Ok, Sounds good, I overlooked it.

dev1644 (approved):

SamAg19 (approved):

SkandaBhat (approved):

PR #103, opened 2021-04-20
solved compilation warnings

adi44:
Some compilation warnings were there. I removed the public keyword from ACL.sol and SchellingCoin.sol. In version 0.8.0 compiler it was showing that the visibility of constructors are ignored. Also I removed the unused constants from StakeManager.sol


abhishekvispute:
As this PR focuses on solving warnings, can we add an SPDX license identifier as well, would solve all warnings.

adi44:
@abhishekvispute Yeah, maybe in today's discussion we can talk about what license to use

adi44:
Closing this PR since this commit is included in #106

PR #110, opened 2021-04-21
Removed the schellingCoin deployment code

adi44:
Currently, I have commented on the two lines which execute the deployment of SchellingCoin.sol. So if in the future we want to deploy again we can remove the comment.


SkandaBhat:
@adi44 Remove instead of commenting, we can always bring it back using git history.

dev1644:
> @adi44 Remove instead of commenting, we can always bring it back using git history.

Agreed with @SkandaBhat here, that's why the `Schelling coin` deployment was added to the last so it can be removed with out changing the migration process much.

@adi44 You can delete `Schelling coin` deployment & `9_delpoy_schelling_coin.js` file.

migrations/deploy_all.js
```diff
@@ -20,7 +20,7 @@ async function main() {
   await voteManagerMigration();
   await delegatorMigration();
   await faucetMigration();
-  await schellingCoinMigration();
+  //await schellingCoinMigration();
```
dev1644, line 23: @adi44 You can delete `Schelling coin` deployment here & also `9_delpoy_schelling_coin.js` file in `/migrations/src`.

SkandaBhat (dismissed):
LGTM

migrations/deploy_all.js
adi44: @dev1644 I have done the needful

dev1644:
Hi @adi44 , currently changes are not too big and also have same objective.
It would be better if you can roll (squash) them into one commit instead of three.

adi44:
@dev1644 I have done the needful

dev1644 (approved):

dev1644 (approved):

PR #114, opened 2021-04-26
Updated the import from OpenZeppelin to ACL.sol

adi44:


dev1644 (commented):
Let's add one test case here to check whether admin role is properly added or not, this test case can be added in all contract test files which inherits `ACL contract`. @adi44

dev1644 (approved):

SkandaBhat (approved):

PR #134, opened 2021-05-24
hotfix, fixed securtiy issues with init function.

### LINKED ISSUE #113: `init` function is poorly designed & can be accessed by anyone for malicious intent
**Describe the bug**
Some of the contracts have `init` function to initialize the state & external contract address(dependencies) of the contracts.
But this function is not restricted by access and can be reused again & again maliciously.

**To Reproduce**
Steps to reproduce the behaviour:
1. Invoke `init` function in any contract (if the function is available).
2. Set the contract addresses to malicious arbitrary external contracts.
3. Function behaviour will be unpredictable & will be exploited eventually.

**Expected behaviour**
`init` function should have limited access & can't be reused, this should work like a constructor.

P.S. Analyze which functions shouldn't be allowed to invoked before `init` is invoked!

dev1644:
Removed init function & replaced by initialize function which is being used to set up external contract addresses after deployment.

initialize function will have the following properties -
- it will be invokable only once (like a constructor), hence restricted re-setup of external contract addresses like it was for the init function
- only be invokable by account with admin role
- functions which depends upon initialization of the contract will revert if the contract is not initialized 

other misc changes -
- more tests on initialize functionality
- JobManager contract doesn't need to initialize functionality, so the init function is removed entirely
- updated migration suite
- removed init functions from interfaces
- minor refactoring
-  updated listing config, deploy script & env files.

Resolves #113


.env.ganache
```diff
@@ -9,7 +9,7 @@ PROVIDER_HOST=127.0.0.1
 PROVIDER_PORT=8545
 
 # Schelling Coin Address (if want to re-use already deployed instance)
-SCHELLING_COIN_ADDRESS=0xB007d2B950467c397846e7797675B76128f8B2C4
+SCHELLING_COIN_ADDRESS=
```
hrishikeshio, line 12: ? Unrelated change?

contracts/Core/BlockManager.sol
```diff
@@ -50,17 +51,17 @@ contract BlockManager is ACL, BlockStorage {
...
+        address _stateManager,
+        address _voteManager,
+        address _jobManager
+    ) external initializer onlyRole(DEFAULT_ADMIN_ROLE)
     {
-        stakeManager = IStakeManager(_stakeManagerAddress);
-        stateManager = IStateManager(_stateManagerAddress);
-        voteManager = IVoteManager(_voteManagerAddress);
-        jobManager = IJobManager(_jobManagerAddress);
+        stakeManager = IStakeManager(_stakeManager);
```
hrishikeshio, line 61: why was _stakeManagerAddress renamed to _stakeManager
The "Address" part should be there. stakeManager and _stakeManager are different data types.

hrishikeshio: same for other variables

hrishikeshio (changes requested):
LGTM. Minor changes requested.

.env.ganache
dev1644: Whenever we deploy on ganache(for local-testing), we don't have existing ` SCHELLING_COIN_ADDRESS` & won't be reusing it. So Ganache deployment should be from scratch but if someone wants redeployment, he can easily modify the `.env.ganache` according to their needs.

contracts/Core/BlockManager.sol
dev1644: >  The "Address" part should be there. stakeManager and _stakeManager are different data types.

Since typecasting it to the same type is being done immediately, I don't think it's an issue. But there are other reasons why I renamed this - 

- According to solidity's default code-style convention, the underscore prepended with identifier only if there exists a storage variable with that name or a function/state variable is private. (This guide is highly subjective, I am referencing a default recommendation).

- Yes initially the type of `initialize` args are different but imminent type-casting is done.

Solution - Either rename the args to `stakeManagerAddress` or `_stakeManager`, I prefer latter.

hrishikeshio: I would prefer stakeManagerAddress since it is more descriptive and less confusing. Please avoid changes unrelated to the issue in the future.

abhishekvispute (approved):

hrishikeshio (approved):
lgtm

Issue #104, opened 2021-04-20
Fetch grace period from Constants and make it consistent


abhishekvispute:
Currently in the grace period value is hardcoded, we have to make it constant so we can expose it as a governance parameter
```
  function calculateInactivityPenalties(uint256 epochs, uint256 stakeValue) public pure returns(uint256) {
        if (epochs < 10) {
            return(stakeValue);
        }
    }
```

Plus definition of grace period is not consitent

At line https://github.com/razor-network/contracts/blob/8f59a1e78f886bedb433c1fcd3592af1296753f6/contracts/Core/StakeManager.sol#L374
 we are calculating no of penalizing epochs as 
Current Epoch - Last Active and passing it to CalcalateInactivityPenalties()
https://github.com/razor-network/contracts/blob/8f59a1e78f886bedb433c1fcd3592af1296753f6/contracts/Core/StakeManager.sol#L377

The thing is this does not represent no of inactive epochs Current Epoch - Last Active -1 does 
We have kept this in mind while calculating the inactivity penalty as seen in 
https://github.com/razor-network/contracts/blob/8f59a1e78f886bedb433c1fcd3592af1296753f6/contracts/Core/StakeManager.sol#L344

But for the grace period, we have not
https://github.com/razor-network/contracts/blob/8f59a1e78f886bedb433c1fcd3592af1296753f6/contracts/Core/StakeManager.sol#L340
Plus 
10 is not grace here 
8 is , If we look closely 
8 is the maximum no of inactive epochs we are allowing 
We have to make this more readable

PR #137, opened 2021-05-29
Fetching grace period from parameters and making its use readable

abhishekvispute:
Fixes https://github.com/razor-network/contracts/issues/104

Hi Guys,
This is small PR.
Firstly grace period is defined as constant and fetched from there so in the future we can expose it as a governance parameter.
Secondly 
rather than passing "Current Epoch - Last Active", 
now we are passing "Current Epoch - Last Active-1" to calculateInactivePenalties()
Which actually represent no of inactive epochs
Please refer to issue https://github.com/razor-network/contracts/issues/104 for more details

**Please note
This doesn't change any earlier behavior, as confirmed by tests as well.**

Rather than having
X < 10
first, we went 
X-1 < 9
then
X-1 <=8

That's all!

Here X is Current Epoch - Last Active

Signed-off-by: abhishekvispute <abhivispute33@gmail.com>

### DIFF
--- a/contracts/Core/Parameters.sol
+++ b/contracts/Core/Parameters.sol
@@ -19,7 +19,7 @@ contract Parameters is IParameters, ACL {
     uint256 public override epochLength = 300;
     uint256 public override numStates = 4;
     uint256 public override exposureDenominator = 1000;
-
+    uint256 public override gracePeriod = 8;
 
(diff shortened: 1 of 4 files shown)


SkandaBhat (approved):
lgtm

SamAg19 (approved):

PR #138, opened 2021-05-30
Dissolved Constant library, created separate Parameters contract.

### LINKED ISSUE #75: Dissolve Constants library, declare constants into contract if needed.
This issue is about to dissolve `Constants.sol` library, declare constants into the contract if needed.

This has following benefits -:

- As constants are inlined in Solidity, we would be saving some execution cost too. When we fetch constants from Library it uses delegate call which adds up the cost of tx. (~2000 gas)
- Almost all of the constants declared in `Constnats.sol` are exclusive to a single contract, it will improve code readability too.

dev1644:
Removed Constants library & StateManger Contract, implemented Constants contract, adjusted migration functionality accordingly.

- Removed Constants library & StateManager Contract, the functionality of StateManager has been implemented into new Constants Contract
- Constants Contract have Access Control, Admin can change the values of variables in the contract
- StateManger contract's functionality has been incorporated into Constants Contact itself.
- New tests for new Constants Contract
- Refactored tests suite accordingly
- Removed Constants library, StateManager contract migration
-  Added Constants Contract migration

Resolves #75


contracts/Core/StateManager.sol
```diff
@@ -1,17 +0,0 @@
-// SPDX-License-Identifier: UNLICENSED
-pragma solidity ^0.8.0;
-
-import "../lib/Constants.sol";
-
-
-contract StateManager {
```
abhishekvispute, line 7: I think we should keep State Manager as it was before. There are two reasons. First :  Constants contract should only have constants which are static. so its clear around, Second : Architecture remains readable. I am not sure what is benefit of keeping it in constants ? @dev1644 Please let me know if there are any ?

contracts/Core/Constants.sol
```diff
@@ -0,0 +1,123 @@
...
+    // keccak256("JOB_CONFIRMER_ROLE")
+    bytes32 constant private _JOB_CONFIRMER_HASH = 0xbe7b58e17bf6adaa0f209cd0db8b128282fc68a42f2dd649b4d8ea579f1b078f;
+    
+    // keccak256("STAKER_ACTIVITY_UPDATER_ROLE")
+    bytes32 constant private _STAKER_ACTIVITY_UPDATER_HASH = 0x4cd3070aaa07d03ab33731cbabd0cb27eb9e074a9430ad006c96941d71b77ece;
+    
+    // keccak256("STAKE_MODIFIER_ROLE")
+    bytes32 constant private _STAKE_MODIFIER_HASH = 0xdbaaaff2c3744aa215ebd99971829e1c1b728703a0bf252f96685d29011fc804;
+
+    function setCommit(uint32 _commit) external onlyRole(DEFAULT_ADMIN_ROLE) {
```
abhishekvispute, line 42: Hey, I don't think we need setters for state identifiers. Commit, Reveal, Propose and Dispute
These constants would never change

contracts/Core/JobManager.sol
```diff
@@ -50,7 +48,7 @@ contract JobManager is ACL, JobStorage {
     ) external payable 
     {
         numJobs = numJobs + 1;
-        uint256 epoch = stateManager.getEpoch();
+        uint256 epoch = constants.getEpoch();
```
abhishekvispute, line 51: see here with ```constants.getEpoch();``` I get intuition that epoch is some constant that you are returning.
While with ```stateManager.getEpoch()``` there is no such confusion. 
As the state is an entity **I think we should have StateManager like earlier**

contracts/lib/Random.sol
```diff
@@ -1,29 +1,27 @@
...
+        bytes32 hash = prngHash(numBlocks, seed, epochLength);
         uint256 sum = uint256(hash);
         return(sum%max);
     }
 
     // pseudo random hash generator based on block hashes.
-    function prngHash(uint8 numBlocks, bytes32 seed) public view returns(bytes32) {
-        bytes32 sum = blockHashes(numBlocks);
+    function prngHash(uint8 numBlocks, bytes32 seed, uint256 epochLength) public view returns(bytes32) {
+        bytes32 sum = blockHashes(numBlocks, epochLength);
```
abhishekvispute, line 16: Hey why are we adding epochLength as new parameter herr

contracts/Core/Constants.sol
```diff
@@ -0,0 +1,123 @@
...
+
+    function setPenaltyNotRevealNum(uint256 _penaltyNotRevealNumerator) external onlyRole(DEFAULT_ADMIN_ROLE) { 
+        penaltyNotRevealNum = _penaltyNotRevealNumerator;
+    }
+
+    function setPenaltyNotRevealDeom(uint256 _penaltyNotRevealDenom) external onlyRole(DEFAULT_ADMIN_ROLE) { 
+        penaltyNotRevealDenom = _penaltyNotRevealDenom;
+    }
+
+    function setUnstakeLockPeriod(uint256 _unstakeLockPeriod) external onlyRole(DEFAULT_ADMIN_ROLE) {
```
abhishekvispute, line 66: See rather than ```onlyRole(DEFAULT_ADMIN_ROLE)```
Can we create one role like say "Governer" 
The thing is in near future we want DAO to govern these values, so it would be better if we have a dedicated role.
As of now, we can assign this Governer role to Admin.

contracts/Core/StateManager.sol
dev1644: > I think we should keep State Manager as it was before

`StateManager` only had `getEpoch` & `getState` functions, these functions heavily relied upon data from particular values (that are provided by Constants lib/contract). 

Every external calls from a contract to another has some fixed stipend to them (excluding computation cost of that function).
Since only functionality of `StateManager` was just to compute & fetch current Epoch & State (which in turn relies upon values provided by Constants lib/contract), it makes sense to have Constants contract/library to encapsulate `getEpoch & getState` functionality as well in a single contract (`StateManager` offers nothing to have it in its own external contract, at least for now) from gas expenses & complexity perspective.

> I am not sure what is benefit of keeping it in constants ?

Benefits of keeping it in constants is to save few thousand gas (Need more analysis to calculate exact amount).

contracts/Core/Constants.sol
dev1644: Agreed, I just put these functions to confirm whether we need to modify them or not.

If they aren't subject to change, these will be made `constants` like Role parameters.

contracts/Core/JobManager.sol
dev1644: I don't think its justifiable to spend more gas (few thousands) & redundant external calls (which can be trimmed) here.

If we are talking about intuition, then I think `Constants`  name is also wrong since contract will have values that can be modified by admin at some point in future.

Need to think of a suitable name, Its clear `Constants` doesn't work :)

contracts/Core/Constants.sol
dev1644: >  Can we create one role like say "Governer"

Sounds good to me, but for now I avoided it, why? Because currently there is not a clear roadmap how `DAO`  will be implemented, I would rather avoid implementing something that is not quite on the spec yet.

Though I agree with you, `Governer` role can be added in near future.

abhishekvispute: Okay I agree, maybe we should create issue for the same

contracts/Core/StateManager.sol
abhishekvispute: Yes, my bad ! lower gas consumption should be our top-most priority, naming, architecture comes later
Okay 
Can we do it like this maybe
Define those two constants: numStates and epochLength in State Manager only
as numStates is only being used my State Manager 
whereas though epochLength is being used by BlockManager as well, there we have an instance of State Manager, so we can access it if we define epochLength in State Manager

**Please check I think this would solve my arguments and yours as well.** 

State Manager  - All parameters related to state
> numStates
> epochLength
> commit id (adding these here would make more sense )
> reveal id
> propose id
> dispute id

But again one can argue that we want all constants in one place :)
To solve this 
From constants, we can have an external call to StateManager to set up these values, 
So our governing remains in constants.sol only
 
What do you think ?

contracts/Core/JobManager.sol
abhishekvispute: Yep

abhishekvispute: maybe Governance.sol ?

dev1644: > maybe Governance.sol ?

Hmm, we don't have governance logic/mechanism in the contract yet, so it would be misleading IMO. WDYT @hrishikeshio @SkandaBhat @SamAg19 ?

contracts/Core/StateManager.sol
dev1644: > Define those two constants: numstates and epochLength in State Manager only 

Separating constants (just referring to them as constants, it's clear these params are not constants) into a different contract doesn't make sense to me, some of the constants will be in X contract & some of them in Y contract. Moreover, the current implementation of `StateManager` is more of a `StateViewer/Getter` not of a `Manager`.

> Define those two constants: numStates and epochLength in State Manager only as numStates is only being used my State Manager anyway

Yes, `numStates` is being used in `StateManager` only, in `getState` function but in turn, this function is being used much time in multiple contracts (`BlockManager`, `VoteManager`, `StakeManager`).

I understand that some of the parameters in constants lib/contract are never gonna change (this need to be confirmed again though), once confirmed I think it would be better if these immutable constants parameters will exist in all of the contracts with the `constant` type (it will increase one time-deployment gas but will save runtime gas during tx).  This needs to discussed first.

>  Have set for epochLength as well from here by external call

Again, I am not seeing the incentive for using two external calls to set/get a parameter when it can be done with one call only. 


Regarding `never changing constants parameters`, I think it's better to dissolve `Structs` lib too & have a common `Storage` contract which will be inherited by another contract, `never changing constants parameters` will exists in `Storage` contract as of `constant type`.

abhishekvispute: Okay looks good to me as of now
Let's Merge this now 
And we will create issues with milestone v2 for open points CC: @hrishikeshio @SkandaBhat @SamAg19 
Can you please create @dev1644?

**Issue 1**
Define never changing constants in contracts only to avoid runtime external calls.

**Issue 2**
 Actually, I am not totally clear on following, Please explain it issue

> Regarding never changing constants parameters, I think it's better to dissolve Structs lib too & have a common Storage contract which will be inherited by another contract, never changing constants parameters will exists in Storage contract as of constant type.

**Issue3**
Replace Admin role with governer role in constants.sol

abhishekvispute (approved):

contracts/Core/JobManager.sol
SkandaBhat: Agreed ^ we can have parameters.sol maybe.

contracts/Core/Constants.sol
SkandaBhat: Admin role works for now.

contracts/Core/Constants.sol
SkandaBhat: Agreed, we dont need this

contracts/Core/BlockManager.sol
```diff
@@ -278,10 +277,10 @@ contract BlockManager is ACL, BlockStorage {
     {   
         // generating pseudo random number (range 0..(totalstake - 1)), add (+1) to the result,
         // since prng returns 0 to max-1 and staker start from 1
-        if ((Random.prng(10, stakeManager.getNumStakers(), keccak256(abi.encode(iteration)))+(1)) != stakerId) {
+        if ((Random.prng(10, stakeManager.getNumStakers(), keccak256(abi.encode(iteration)), constants.epochLength())+(1)) != stakerId) {
```
SkandaBhat, line 280: Why was this done?

contracts/Core/BlockManager.sol
```diff
@@ -278,10 +277,10 @@ contract BlockManager is ACL, BlockStorage {
...
             return false;
         }
-        bytes32 randHash = Random.prngHash(10, keccak256(abi.encode(stakerId, iteration)));
+        bytes32 randHash = Random.prngHash(10, keccak256(abi.encode(stakerId, iteration)), constants.epochLength());
```
SkandaBhat, line 283: Why was this done?

contracts/Core/JobManager.sol
SamAg19: tbh, Parameters.sol sounds good

contracts/Core/BlockManager.sol
abhishekvispute: Its because Random.prng() method is changed.
It didn't have epoch length as function param before
Why @dev1644 added it I think
Is without this passing, we would have to inherit constants.sol in random as well, because prng uses it 

Old
https://github.com/razor-network/contracts/blob/8f59a1e78f886bedb433c1fcd3592af1296753f6/contracts/lib/Random.sol#L26

contracts/Core/BlockManager.sol
abhishekvispute: Same reason as of above

contracts/Core/BlockManager.sol
dev1644: Correct @abhishekvispute , thus reducing the size of  `Random Library` & minimizing any redundant call.

contracts/Core/StateManager.sol
SkandaBhat: Issue 1 and 2 already exist. We need to create issues for them and track.

abhishekvispute (approved):
LGTM

abhishekvispute:
Issues reported by @dev1644 
To be solved in relation to this PR
https://github.com/razor-network/contracts/issues/143
https://github.com/razor-network/contracts/issues/144

SkandaBhat (approved):

SkandaBhat (approved):

PR #145, opened 2021-06-03
refactored_epoch_length_to_300

### LINKED ISSUE #139: Change the epoch length from 40 to 300 for Matic deployment
The blocktime for goerli is 15 seconds whereas for matic it's 2 seconds. The current epoch length is set according to the goerli but this epoch length doesn't work for matic. It does not sync with the states due to which transactions are reverted

the ideal way should be to change epoch length to 300 before deploying it to matic

adi44:
The blocktime for goerli is 15 seconds whereas for matic it's 2 seconds. The current epoch length is set according to the goerli but this epoch length doesn't work for matic. It does not sync with the states due to which transactions are reverted

in order to match Matic and goerli block time, I have changed the epoch length to 300

Fixes #139


SkandaBhat (approved):
lgtm

abhishekvispute (approved):
LGTM

PR #149, opened 2021-06-04
Upgraded solidity version to 0.8.4

### LINKED ISSUE #132: upgrade compiler version to 0.8.4
hardhat config uses 0.8.0. Current stable solidity version is 0.8.4
use 0.8.4 compiler version in hardhat config and if required, in ci config

adi44:
Currently, we were using the solidity version 0.8.0 but the current stable version is 0.8.4. I have upgraded the solidity version to 0.8.4.

Fixes #132


abhishekvispute (approved):
LGTM

hrishikeshio:
Actually, I think only the hard hat configuration should be changed. We will have to keep updating all solidity files with every minor release which is unnecessary.

Issue #152, opened 2021-06-09
Make Slashing penalty as a governance parameter


abhishekvispute:
**Describe the bug**
As of now slashing penalty is hardcoded, we need to make it a gov parameter and reduce it,
as directly slashing to 0 is too strict.

SkandaBhat:
@GauravJain9 Please pick this up

GauravJain9:
@SkandaBhat  ok

GauravJain9:
Generated a PR for this  @SkandaBhat  
https://github.com/razor-network/contracts/pull/157#issue-667868202

PR #157, opened 2021-06-11
Fixed Isuue Slashing Penalty (#152)

GauravJain9:
I have added a gov param as "percentSlash" by which we can decide and update how much percentage of the staker should be slashed and updated the slash function accordingly all tests have also passed.

### DIFF
--- a/contracts/Core/Parameters.sol
+++ b/contracts/Core/Parameters.sol
@@ -21,6 +21,7 @@ contract Parameters is IParameters, ACL {
     uint256 public override exposureDenominator = 1000;
     uint256 public override gracePeriod = 8;
     uint256 public override aggregationRange = 3;
+    uint256 public override percentSlash = 100;
 
     uint32 constant private _COMMIT = 0;
     uint32 constant private _REVEAL = 1;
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -46,7 +46,7 @@
     "eslint": "^5.16.0",
     "eslint-config-airbnb-base": "^14.2.1",
     "eslint-plugin-import": "^2.22.1",
-    "ethers": "^5.0.31",
+    "ethers": "^5.3.1",
```
SkandaBhat, line 49: Why was this change made?

contracts/Core/Parameters.sol
```diff
@@ -21,68 +21,73 @@ contract Parameters is IParameters, ACL {
...
     uint256 public override aggregationRange = 3;
+    uint256 public override percentSlash = 100;
 
     uint32 constant private _COMMIT = 0;
     uint32 constant private _REVEAL = 1;
     uint32 constant private _PROPOSE = 2;
     uint32 constant private _DISPUTE = 3;
 
     // keccak256("BLOCK_CONFIRMER_ROLE")
-    bytes32 constant private _BLOCK_CONFIRMER_HASH = 0x18797bc7973e1dadee1895be2f1003818e30eae3b0e7a01eb9b2e66f3ea2771f;
```
abhishekvispute, line 31: Please dont commit linting changes

abhishekvispute: Only addition should be setSlashPenalty()

SkandaBhat (changes requested):
There are too many formatting changes which make it difficult to review the PR. Please remove them and include just the changes that are required.

contracts/Core/StakeManager.sol
```diff
@@ -297,12 +297,13 @@ contract StakeManager is Initializable, ACL, StakeStorage {
     /// @param id The ID of the staker who is penalised
     /// @param bountyHunter The address of the bounty hunter
     function slash (uint256 id, address bountyHunter, uint256 epoch) external onlyRole(parameters.getStakeModifierHash()) {
-        uint256 halfStake = stakers[id].stake/(2);
-        _setStakerStake(id, 0, "Slashed", epoch);
-        if (halfStake > 1) {
-            require(sch.transfer(bountyHunter, halfStake), "failed to transfer bounty");
+      uint256 halfStake = ((stakers[id].stake)*(parameters.percentSlash()))/(2*100);
+      uint256 Stake =  stakers[id].stake - ((stakers[id].stake*parameters.percentSlash())/100);
```
abhishekvispute, line 301: I think percentSlash is not clear
slashPenalty ?

abhishekvispute: Again here only addition should be changing to governance param, please remove lint changes

package.json
abhishekvispute: yes

abhishekvispute (changes requested):

contracts/Core/StakeManager.sol
GauravJain9: @abhishekvispute it is how much percent of the stake should we slash of the staker so it can be percentSlashPenalty ?

package.json
abhishekvispute: yes

PR #159, opened 2021-06-14
Made SlashingPenalty a Governance Parameter

### LINKED ISSUE #152: Make Slashing penalty as a governance parameter

GauravJain9:
Fixes #152
Added a gov param as "percentSlashPenalty" by which we can decide and update how much percentage of the staker should be slashed and updated the slash function accordingly.


GauravJain9:
@SkandaBhat @abhishekvispute  @SamAg19 Look into this PR

contracts/Core/StakeManager.sol
```diff
@@ -292,17 +292,18 @@ contract StakeManager is Initializable, ACL, StakeStorage {
...
     function slash (uint256 id, address bountyHunter, uint256 epoch) external onlyRole(parameters.getStakeModifierHash()) {
-        uint256 halfStake = stakers[id].stake/(2);
-        _setStakerStake(id, 0, "Slashed", epoch);
-        if (halfStake > 1) {
-            require(sch.transfer(bountyHunter, halfStake), "failed to transfer bounty");
-        }
-    } 
+      uint256 bountyReward = ((stakers[id].stake)*(parameters.percentSlashPenalty()))/(2*100);
+      uint256 Stake =  stakers[id].stake - ((stakers[id].stake*parameters.percentSlashPenalty())/100);
+      if (bountyReward > 1) {
```
abhishekvispute, line 302: IMO We don't need this if, what it solves I am not clear
There are only two corner cases it excludes

bountyReward ==1 > I am not sure why we are not awarding bounty in this case
bountyReward ==0  > Even if its zero, sch.transfer wont transfer anything 

What do you guys say?
Maybe we should create diff issue and sep PR for same

GauravJain9: if (bountyReward > 0 && bountReward<1)
I think may be it is handling that @abhishekvispute

SamAg19: I agree with @abhishekvispute here. I would also suggest we remove that if statement

GauravJain9:
ok will remove this @SamAg19 @abhishekvispute

contracts/Core/StakeManager.sol
```diff
@@ -292,17 +292,18 @@ contract StakeManager is Initializable, ACL, StakeStorage {
...
     /// @param id The ID of the staker who is penalised
     /// @param bountyHunter The address of the bounty hunter
     function slash (uint256 id, address bountyHunter, uint256 epoch) external onlyRole(parameters.getStakeModifierHash()) {
-        uint256 halfStake = stakers[id].stake/(2);
-        _setStakerStake(id, 0, "Slashed", epoch);
-        if (halfStake > 1) {
-            require(sch.transfer(bountyHunter, halfStake), "failed to transfer bounty");
-        }
-    } 
+      uint256 bountyReward = ((stakers[id].stake)*(parameters.percentSlashPenalty()))/(2*100);
```
SkandaBhat, line 300: We are doing similar computation twice here. Can this be optimised?

contracts/Core/StakeManager.sol
SkandaBhat: +1

SkandaBhat (changes requested):

contracts/Core/StakeManager.sol
SkandaBhat: Recorded this in #162

contracts/Core/StakeManager.sol
abhishekvispute: Yes first penalty should be calculated 
Then bountyReward  = penalty / 2

contracts/Core/StakeManager.sol
SkandaBhat: if (bountyReward = 0) return;
// go ahead

GauravJain9: will update this at once after updating all the test cases regarding this.

PR #165, opened 2021-06-18
Integrated NPM Package.

### LINKED ISSUE #148: Publish npm package
- Compile contracts
- publish ABIs to npm

We will not be adding the address in the package.

Also, go-cli needs to generate bindings independently.

adi44:
Currently, user needs to import the contracts through the deployed folder. Now instead anyone can use NPM Package of contracts ABI and use it for integration.
Fixes #148


adi44:
conflicts were there

PR #167, opened 2021-06-22
finalizeDispute edge testCase

### LINKED ISSUE #151: finalizeDispute edge case
**Describe the bug**
If there is only one block and that is an invalid block, then the next epoch will have problems in stakeManager.

Lets not penalise in this case.

GaurShekhar:
resolves #151

Description:

Created a new staker and asserted it's stake to check for peanlization and rewards.


test/BlockManager.js
```diff
@@ -446,19 +494,11 @@ describe('BlockManager', function () {
         iteration,
         biggestStakerId);
       const proposedBlock = await blockManager.proposedBlocks(epoch, 0);
-      assertBNEqual(proposedBlock.proposerId, toBigNumber('4'), 'incorrect proposalID');
+      assertBNEqual(proposedBlock.proposerId, toBigNumber('5'), 'incorrect proposalID');
 
       await mineToNextState();
 
       const sortedVotes = [toBigNumber('20000')];
-      const {
```
abhishekvispute, line 454: Hey @GaurShekhar, Why we are removing this

test/BlockManager.js
```diff
@@ -395,13 +417,16 @@ describe('BlockManager', function () {
       assert((await proposedBlock.valid) === false);
     });
 
-    it('no block should be confirmed in the previous epoch', async function () {
+    it('no block should be confirmed in the previous epoch and no peanalization for one block which is invalid', async function () {
```
abhishekvispute, line 420: TestCase looks good. 
It does check our req. Can we rephrase this better maybe ? Like If No Block is valid in previous epoch, Stakers shouldnt be penalised. Its your call @GaurShekhar

test/BlockManager.js
```diff
@@ -413,6 +438,19 @@ describe('BlockManager', function () {
 
       await voteManager.connect(signers[19]).commit(epoch, commitment1);
 
+      const votes2 = [1000, 2000, 3000, 4000, 5000, 6000, 7000, 8000, 9000];
+      const tree2 = merkle('keccak256').sync(votes2);
+
+      const root2 = tree2.root();
+      const commitment2 = utils.solidityKeccak256(
```
abhishekvispute, line 445: Why 2 stakers are needed in this testcase, Here if I understand both 8 and 19 are participating, is it needed ?

test/BlockManager.js
```diff
@@ -122,6 +125,17 @@ describe('BlockManager', function () {
 
       await voteManager.connect(signers[6]).commit(epoch, commitment2);
 
+      const votes3 = [100, 200, 300, 400, 500, 600, 700, 800, 900];
```
abhishekvispute, line 128: Why do we need for your new staker : 8 to participate here ?

test/BlockManager.js
GaurShekhar: Initially we were calculating the median and everything as we were comparing it with the afterDisputeReset but in this test case we are resetting the values so I have modified the test case to check the equality to it's resetting value rather than the calculated value.

test/BlockManager.js
GaurShekhar: Ok will change the test case name.

test/BlockManager.js
GaurShekhar: So for testing this particular edge testCase I decided to add the new staker which will only participate in commit and reveal state and I have commited and revealed it two times  before testing it's stake in 'no block should be confirmed in previous epoch' test case .

test/BlockManager.js
GaurShekhar: So i hvae only updated the previous test case in which 19th staker was partcipating earlier as it also has to propose a block in nest testCase. The new 8th staker is participating for commit and reveal state only

test/BlockManager.js
abhishekvispute: Yep adding new staker is fine, but why we need for them to commit and reveal here, doesnt it make this testcase needlessly bulky ?

test/BlockManager.js
abhishekvispute: oh okay 
can we move this 19th staker part to the next test case then, as that's where it's needed
Ideally, a test case should have only the things concerned with that testcase.
So when we read it it's clear.

test/BlockManager.js
GaurShekhar: Yes

test/BlockManager.js
GaurShekhar: Ok cool, will do that.

abhishekvispute (approved):
LGTM

SkandaBhat (approved):

PR #169, opened 2021-06-22
Admin Access

### LINKED ISSUE #86: Admin access for Job URLs
Since APIs are centralised and failure prone in nature, we need a mechanism to allow for modification and deletion of URLs.

We need to have admin access around job URLs, to allow for addition, modification and deletion.

Deleting jobs, hiding jobs etc are immediate solutions that are not scalable and don't solve the problem.

**Additional context**
We should use snapshot.org for community opinion on this.

SamAg19:
Fixes #86 

Creation, updation or deletion of jobs and collections is possible and can only be done by the address that has been assigned the role "ASSET_CREATOR_ROLE".

Three new functions are added to the AssetManager.sol that allow us to update or delete assets:-

  1. **updateJob(): You can update the URL and selector of the jobID specified.**
  2. **updateAssetStatus(): You can update activity status of the assetID specified. If an asset wants to be deleted, this function needs to be called**
  3. **removeJobFromCollection(): You can remove jobID, if exists, from the collection specifed by the address.**


contracts/Core/AssetManager.sol
```diff
@@ -78,7 +93,10 @@ contract AssetManager is ACL, AssetStorage {
         string calldata selector, 
         string calldata name, 
         bool repeat
-    ) external payable {
+    ) 
+        external 
+        onlyRole(parameters.getAssetCreatorHash())
```
abhishekvispute, line 98: By Admin Access to Jobs, do we mean that only admin can create jobs ?

contracts/Core/AssetManager.sol
```diff
@@ -56,8 +66,6 @@ contract AssetManager is ACL, AssetStorage {
         string name,
         uint32 aggregationMethod,
         uint256[] jobIDs,
-        address creator,
-        uint256 credit,
```
abhishekvispute, line 60: Earlier in job we had credit as param, which I suppsose was for reward for that job, but I didnt find it being transfered anywhere, Was it being used ?

contracts/Core/AssetManager.sol
```diff
@@ -102,14 +118,87 @@ contract AssetManager is ACL, AssetStorage {
...
+        url,
+        selector,
+        block.timestamp
+        );
+
+    }
+
+    function updateAssetStatus(
+        uint256 id,
+        uint256 activationMethod
```
abhishekvispute, line 155: Can we have bool as type directly here, and can we have variable name like isActive ?

contracts/Core/interface/IParameters.sol
```diff
@@ -29,5 +29,6 @@ interface IParameters {
     function getStakeModifierHash() external view returns(bytes32);
     function getStakerActivityUpdaterHash() external view returns(bytes32);
     function getRewardModifierHash() external view returns(bytes32);
+    function getAssetCreatorHash() external view returns (bytes32);
```
abhishekvispute, line 32: As we have update now, does AssetModifier make more sense. AssetCreator as a role is also fine, its your call

contracts/Core/AssetManager.sol
SamAg19: Yes, only admin can create a job/collection

contracts/Core/AssetManager.sol
SamAg19: That functionality isnt being used anywhere and credit was there as fee a user has to pay to create a job but since this can only be called by the admin only, there is no point in having credit and creator.

contracts/Core/AssetManager.sol
SamAg19: Alright, bool can be used here also
Do you want isActive to replace activationMethod or the active variable in the job and collection structs?

contracts/Core/interface/IParameters.sol
SamAg19: It does both so i was kind of confused what to name the role. Maybe AssetCreatorModifier as the address with this role is both creating and modifying assets?

contracts/Core/AssetManager.sol
abhishekvispute: yep

contracts/Core/AssetManager.sol
abhishekvispute: Hmm, my suggestion was for activationMethod only, active in structs should be fine IMO

contracts/Core/interface/IParameters.sol
abhishekvispute: Hmm dont know :joy:, Lets have AssetCreator only then.

contracts/Core/AssetManager.sol
```diff
@@ -31,9 +29,22 @@ contract AssetManager is ACL, AssetStorage {
         string selector,
         string name,
         bool repeat,
-        address creator,
```
SkandaBhat, line 34: IMO creator is required since admin may change. Any specific reason this was removed?

contracts/Core/interface/IParameters.sol
SamAg19: Alright

contracts/Core/AssetManager.sol
SamAg19: Removed it as the admin can only create jobs/collection but makes sense to bring it back as the admin might change.

contracts/Core/AssetManager.sol
```diff
@@ -102,14 +118,87 @@ contract AssetManager is ACL, AssetStorage {
...
+    function updateJob(
+        uint256 jobID,
+        string calldata url,
+        string calldata selector
+    ) 
+        external 
+        onlyRole(parameters.getAssetCreatorHash()) 
+    {
+        require(jobs[jobID].assetType==uint256(assetTypes.Job),"Job ID not present");
+        require(jobs[jobID].active, "Job has been inactivated");
```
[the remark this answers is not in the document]
SamAg19: Alright, will make the change.

contracts/Core/AssetManager.sol
```diff
@@ -102,14 +118,87 @@ contract AssetManager is ACL, AssetStorage {
...
+
+    function updateAssetStatus(
+        uint256 id,
+        uint256 activationMethod
+    ) 
+        external 
+        onlyRole(parameters.getAssetCreatorHash()) 
+    {
+        require(id != 0, "ID cannot be 0");
+        require(id<=numAssets,"ID does not exist");
```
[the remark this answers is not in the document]
SamAg19: oh forgot to use the linter. Will rectify this

contracts/Core/AssetManager.sol
SkandaBhat: 👍

hrishikeshio (approved):
lgtm

abhishekvispute (approved):

SkandaBhat (approved):

PR #175, opened 2021-06-28
Made Slashing Penalty a Governance Parameter

### LINKED ISSUE #152: Make Slashing penalty as a governance parameter

GauravJain9:
Fixes #152
Added a gov param as "percentSlashPenalty" by which we can decide and update how much percentage of the staker should be slashed and updated the slash function accordingly.

### DIFF
--- a/contracts/Core/Parameters.sol
+++ b/contracts/Core/Parameters.sol
@@ -11,6 +11,8 @@ contract Parameters is IParameters, ACL {
     // penalty not reveal = 0.01% per epch
     uint256 public override penaltyNotRevealNum = 1;
     uint256 public override penaltyNotRevealDenom = 10000;
+    uint256 public override slashPenaltyNum = 1000;
+    uint256 public override slashPenaltyDenom = 1000;
 
     uint256 public override minStake = 100 * (10 ** 18);
     uint256 public override withdrawLockPeriod = 1;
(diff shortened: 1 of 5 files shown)


GauravJain9:
@abhishekvispute it's working actually signers[19] didn't had funds so transferred it now shall I keep the tokenamount passed as inintial weights in test case manually or shall I pass it by taking value from VoteManager like for eg. "[tokenAmount('420000'), tokenAmount('18000')] // initial weights" to "[await voteManager.getVoteWeight(epoch, 1, sortedVotes[0])] // initial weights" .

abhishekvispute:
Hey @GauravJain9 
How it was working before in this case then if signers[19] never had those funds?

On keeping it manually or from contract 
First check if the calculated value in your test manyal one is equal to what we get from contract
Then in next test it we can pass any of them

GauravJain9:
Hey @abhishekvispute 
Previously it working because in test case of finalize dispute signers[19] got slashpenalty/2 as reward by slashing signers[5] and the amount signers[19] received was i think tokenAmount('210000') which is far greater than what signers[19] is staking in the below test cases which is tokenAmount('19000') so that's why it was working then.

On keeping it manually or from contract
I have checked in every places by manually and from contract in both the cases same values are being calculated so we can pass any of them either in all or in selected test cases as per requirement. So as of now I haven't changed I have kept it manually again since the issue is resolved and it actually had nothing to do with this.

abhishekvispute (approved):
LGTM

SkandaBhat (approved):

PR #186, opened 2021-07-06
resolved-todo for Parameters.sol

### LINKED ISSUE #172: Analyze what can be the maximum value (if needs to be reset) of these params.
Currently, we have set the value of the parameter, but we haven't decided on what could be the maximum value for every constant. for which we need to do some analysis.

adi44:
there was a to-do comment on Parameters. sol to decide the Maximum Value of all the constants. But as per discussion, we do not need to have maximum values for constants. So I have removed that todo comment from Parameters.sol
fixes #172

### DIFF
--- a/contracts/Core/Parameters.sol
+++ b/contracts/Core/Parameters.sol
@@ -6,7 +6,6 @@ import "./ACL.sol";
 
 contract Parameters is IParameters, ACL {
 
-    // TODO: Analyze what can be the maximum value (if needs to be reset) of these params.
     // constant type can be readjusted to some smaller type than uint256 for saving gas (storage variable packing).
     // penalty not reveal = 0.01% per epch
     uint256 public override penaltyNotRevealNum = 1;


SkandaBhat:
@adi44 This needs more description.

adi44:
> @adi44 This needs more description.

@SkandaBhat I have added description

abhishekvispute:
Hey,
I think it would be better if we remove all todos in one single PR, rather than having PR for each.
We will remove all of these todos once we resolve all of them, at once.
IMO It will be better for tracking.

What do you think ? @SkandaBhat  @adi44

adi44:
> Hey @adi44,
> Is there any other to-do which we have already taken care of, or which is not needed anymore like this?
> I think it would be better if we have one PR removing all of the dormant to-do at once, rather than having a single for each.
> What do you think?

@abhishekvispute Hey! yeah, I think there is one more TODO which is not needed any more. Yeah, I think it would be better to have it in one PR and link the issue.

PR #187, opened 2021-07-06
resolved minstake todo and enhanced condition for stake > 0

### LINKED ISSUE #171: Write a functionality to choose if staker can propose a block if its stake is below min stake.
**Describe the bug**
Currently staker is chosen using probabilistic method using peasudorandom funcion. But in case if stake is below minimum stake there is no method defined to choose a staker. 

Solution Proposed : Staker tosses a bias coin, where bias = minStake/biggestStake. if its head, staker can propose a block else wait for next iteration. 
https://github.com/razor-network/contracts/blob/1b9fcc1e6204f5a47c69cc1a0929bc7519a1e85b/contracts/Core/BlockManager.sol#L124-L130

adi44:
fixes #171


adi44:
it has previous commit also, which is not needed for this PR

PR #195, opened 2021-07-09
Changed RewardPool assginment as ternary condition.

### LINKED ISSUE #178: Tx is reverted because rewards > reward pool
We have to analyse why rewards is greater than reward pool. Could be rewards are being calculated wrong or could be a deeper issue.

adi44:
https://blog.soliditylang.org/2020/12/16/solidity-v0.8.0-release-announcement/. As per this link evm might revert a transaction when there is underflow or overflow. Based on the ternary condition, the rewardPool - reward will only execute, when rewardPool >= reward else it will return 0. this will prevent underflow conditions. 
fixes #178

### DIFF
--- a/contracts/Core/RewardManager.sol
+++ b/contracts/Core/RewardManager.sol
@@ -171,7 +171,7 @@ contract RewardManager is Initializable, ACL, RewardStorage {
                 );
                 uint256 newStake = thisStaker.stake + reward;
                 uint256 prevRewardPool = rewardPool;
-                rewardPool = rewardPool - (reward);
+                rewardPool = rewardPool >= reward ? rewardPool - reward : 0 ;
                 emit RewardPoolChange(
                     epoch,
                     prevRewardPool,
--- a/test/VoteManager.js
+++ b/test/VoteManager.js
@@ -74,10 +74,14 @@ describe('VoteManager', function () {
         await schellingCoin.transfer(signers[4].address, tokenAmount('19000'));
         await schellingCoin.transfer(signers[5].address, tokenAmount('1000'));
         await schellingCoin.transfer(signers[6].address, tokenAmount('1000'));
+        await schellingCoin.transfer(signers[7].address, tokenAmount('1000'));
+
         await schellingCoin.connect(signers[3]).approve(stakeManager.address, tokenAmount('420000'));
         await schellingCoin.connect(signers[4]).approve(stakeManager.address, tokenAmount('19000'));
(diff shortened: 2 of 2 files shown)


hrishikeshio:
Under what scenario can reward > reward pool. 

I don't think it is possible due to line https://github.com/razor-network/contracts/blob/master/contracts/Core/RewardManager.sol#L158

Also, no tests are provided. This PR can't be merged.

adi44:
@hrishikeshio there might be a scenario where all stakers commit votes in all the epoch for which they will be getting rewards. Since RewardPool amount depends on penalties, if there will be no penalties imposed, then reward pool will remain 0. Like in our case, the stakers we are running on cloud votes in all the epochs, the reward pool keeps on reducing eventually. this might lead to a case where rewardPool becomes less than reward to be given to staker.

hrishikeshio:
> @hrishikeshio there might be a scenario where all stakers commit votes in all the epoch for which they will be getting rewards. Since RewardPool amount depends on penalties, if there will be no penalties imposed since everyone votes as expected, then reward pool will remain 0. Like in our case, the stakers we are running on cloud votes in all the epochs, the reward pool keeps on reducing eventually. this might lead to a case where rewardPool becomes less than reward to be given to staker. I have tested locally by running node .



If rewardPool is 0 reward is also 0 at this line https://github.com/razor-network/contracts/blob/master/contracts/Core/RewardManager.sol#L158

so that case is not possible.

Create another PR with testcase that fails due to this.

adi44:
> > @hrishikeshio there might be a scenario where all stakers commit votes in all the epoch for which they will be getting rewards. Since RewardPool amount depends on penalties, if there will be no penalties imposed since everyone votes as expected, then reward pool will remain 0. Like in our case, the stakers we are running on cloud votes in all the epochs, the reward pool keeps on reducing eventually. this might lead to a case where rewardPool becomes less than reward to be given to staker. I have tested locally by running node .
> 
> If rewardPool is 0 reward is also 0 at this line https://github.com/razor-network/contracts/blob/master/contracts/Core/RewardManager.sol#L158
> 
> so that case is not possible.
> 
> Create another PR with testcase that fails due to this.

Yeah!

adi44:
I have written test case to check if it is able to reveal or not if the reward amount is greater than reward pool, for the earlier implementation, where rewardPool = rewardPool - reward, it reverts the transaction. For more details check PR #199. 
For the current implementation where rewardPool = rewardPool>=reward ? rewardPool - reward : 0, the testcase works.

SkandaBhat:
We need more info around this and this may not solve the issue. Closing this for now.

PR #199, opened 2021-07-12
testcase : RewardPool < Reward

adi44:
Currently, when the reward pool is less than the reward Amount the test case fails,.


hrishikeshio:
You are manually manipulating reward pool, stake, etc. This is not a valid test.

Issue #135, opened 2021-05-25
Assign Collections Randomly


SkandaBhat:
**Is your feature request related to a problem? Please describe.**
We have to optimise merkle proof computation.

**Describe the solution you'd like**
- Assign Jobs to stakers randomly solves this problem.
- Number of jobs assigned to each staker needs to be decided by governance parameter

SkandaBhat:
Is related to #46.

abhishekvispute:
And its back 🚀

SkandaBhat:
This has a dependency on #582

PR #206, opened 2021-07-12
Assign Assets Randomly

abhishekvispute:
Fixes https://github.com/razor-network/contracts/issues/135

Earlier Stakers used to commit and reveal all of the assets, which used to cause more gas in Merkle Proof Verification.
What this PR does is basically optimize this scenario by assigning assets randomly

---

1. Staker commits all

3. Staker with the help of PRNG finds assigned assets.

```
AssignedAsset(i) = PRNG (BlockHash : K(Last NumBlocks), Max : NumAssets, Seed : K(StakerId + i)) 
0 < i < maxAssetsPerStaker
K(x) = keccak256(x)
```
> Taking BlockHash of Last NumBlocks makes it impossible for staker to know the assigned assetids for them or anyone in commit state, hence they have to be honest.
> 
> PRNG will generate random no between 1 to Max: NumAssets
> This NumAssets, we have to update to NumActiveCollections, @SamAg19, 
and rather than picking from the basket of 1 to n assets, we have to pick from basket[] of 1 to n active assets
 
3. Staker only reveals these assigned assets 

Note: This would impact CLI as well , CC : @ashish10677 , @SkandaBhat 

---
Now, where did this change had impact other than reveal?

**RewardManager**
giveRewards : only give rewards for revealed assets, refactored the code also little, so its structure and of givePenalties remains consistent
givePenalties : only penalize for revealed assets.

**BlockManager**
Removed redundant methods, giveMedians, giveLC and all, as now they are not being used by RewardManager anymore.
Fixes https://github.com/razor-network/contracts/issues/225
finalizeDispute: Added one more param in input to locate the assetId in block , as index in the block is no longer the same to assetId
(As it's not given that all assets are going to be included in a block)

**AssetManager**
Adeded getNumAssets() to interface: required in PRNG calcualution in voteManager 

**Tests** 
All tests were static before, 
Like we had always considered that 10 jobs are going to be revealed
10 jobs are going to be proposed and all.
We had to make everything dynamic to support random assignments,


smercan89 (approved):

contracts/Core/BlockManager.sol
```diff
@@ -67,49 +66,21 @@ contract BlockManager is Initializable, ACL, BlockStorage {
         parameters = IParameters(parametersAddress);
     }
 
-    function getBlock(uint256 epoch) external view returns(Structs.Block memory _block) {
+    function getBlock(uint256 epoch) external view returns(Structs.Block memory) {
         return(blocks[epoch]);
     }
 
-    function getBlockMedians(uint256 epoch) external view returns(uint256[] memory _blockMedians) {
```
abhishekvispute, line 74: These Methods were redundant (we can get everything from the block directly), so have removed them.
Let me know if we should have them.
Or If we should remove them in separate PR.

contracts/Core/RewardManager.sol
```diff
@@ -127,37 +127,37 @@ contract RewardManager is Initializable, ACL, RewardStorage {
...
         if ((epoch - epochLastRevealed) != 1) return;
-        uint256[] memory mediansLastEpoch =
-            blockManager.getBlockMedians(epochLastRevealed);
-        uint256[] memory lowerCutoffsLastEpoch =
-            blockManager.getLowerCutoffs(epochLastRevealed);
-        uint256[] memory higherCutoffsLastEpoch =
-            blockManager.getHigherCutoffs(epochLastRevealed);
 
-        if (lowerCutoffsLastEpoch.length > 0) {
+        Structs.Block memory blockLastEpoch = blockManager.getBlock(
```
abhishekvispute, line 131: This change was made to keep givePenalties and giveRewards structure consistent.

contracts/Core/BlockManager.sol
```diff
@@ -67,49 +67,21 @@ contract BlockManager is Initializable, ACL, BlockStorage {
...
```
abhishekvispute, line 74: These getters were removed as they are no longer being used in giveRewards()
and as their purpose can be easily fulfilled by getBlock

contracts/Core/Parameters.sol
```diff
@@ -23,7 +23,9 @@ contract Parameters is IParameters, ACL {
     uint256 public override aggregationRange = 3;
     uint256 public override withdrawReleasePeriod = 5;
     uint256 public override resetLockPenalty = 1;
-
+    // Note : maxAssetsPerStaker should be less than total no of jobs
```
hrishikeshio, line 26: what happens if it isnt?

contracts/Core/VoteManager.sol
```diff
@@ -110,6 +117,12 @@ contract VoteManager is Initializable, ACL, VoteStorage {
         }
     }
 
+    function isAssetAllotedToStaker(uint256 stakerId, uint256 iteration, uint256 assetId) public view initialized returns (bool)
```
hrishikeshio, line 120: Not sure if the logic is correct. let's clarify on call.

contracts/Core/BlockManager.sol
SkandaBhat: Please create another issue and link it to this PR.

abhishekvispute: sure

contracts/Core/Parameters.sol
abhishekvispute: If it isn't , the protocol doesn't remain efficient 
As there will be more chances of the same job assigned again.
code won't break but

contracts/Core/VoteManager.sol
abhishekvispute: sure

hrishikeshio (approved):

contracts/Core/BlockManager.sol
abhishekvispute: done https://github.com/razor-network/contracts/issues/225

SkandaBhat:
This will be merged once #218 is merged.

contracts/Core/Parameters.sol
GauravJain9: @abhishekvispute  maxAssetsPerStaker will always be less than no. of jobs like if we have 1000 jobs then there can be let's say 50 assets so maxAssets < no. of jobs always. So if a staker is assigned all the assets then also it will be less than the number of jobs right ?

abhishekvispute: Hey @GauravJain9, Assets and Job are same in this context

hrishikeshio (dismissed):

contracts/Core/Parameters.sol
GauravJain9: oh got it thanks @abhishekvispute

hrishikeshio (approved):

SkandaBhat (approved):

PR #212, opened 2021-07-16
updateAssetStatus() functions in opposite way than expected  

### LINKED ISSUE #203: updateAssetStatus() functions in opposite way than expected
**Describe the bug**
calling updateAssetStatus() with isActive = true should activate the job. 

However, it deactivates the job which is counterintuitive.

it should activate the job/collection.

Also, the If statement here is redundant
https://github.com/razor-network/contracts/blob/master/contracts/Core/AssetManager.sol#L171

Can be replaced with simply
```
jobs[id].active = isActive;
```

GauravJain9:
fixes #203 
1) changed function name (updateAsssetStatus() => setAssetStatus())
Reason :- In this function we are setting the status of the assets and directly and not toggling anymore. 

2) Changed the variable name (isActive => assetstatus) 
Reason :-  isActive was being used to check if it's active or not and do the opposite accordingly but now we are updating the active status of asset directly based on the variable passed namely (assetstatus) i.e. whatever status of the asset we want to set.

3) Changed all the test cases in ACL.js and AssetManager.js which were required accordingly , basically all the values passed to the function updateActiveStatus was changed to setActiveStatus and second parameter of being reversed.


test/AssetManager.js
```diff
@@ -158,12 +158,12 @@ describe('AssetManager', function () {
     });
 
     it('should not be able to update activity status of asset if assetID is 0', async function () {
-      const tx = assetManager.updateAssetStatus(0, 0);
```
hrishikeshio, line 161: why is an integer "0" being passed instead of a `bool`?
`bool` should be passed for consistency

hrishikeshio (changes requested):

test/AssetManager.js
SkandaBhat: +1

GauravJain9: replaced from numeric to appropriate boolean values @SkandaBhat @hrishikeshio

hrishikeshio (approved):
lgtm

SkandaBhat (approved):

PR #218, opened 2021-07-22
Influence mechanism

### LINKED ISSUE #168: Penalties are applied on current stake and not stake during previous epoch
**Describe the bug**
The penalties in _givePenalties() function are applied based on the current stake. If a delegator stakes a large sum of tokens before the staker commits, a large amount penalty will be applied.

**To Reproduce**
epoch 1: staker votes incorrect value
epoch 2: Delegator stakes large some of the tokens
epoch 2: staker commits

**Expected behavior**
staker should be penalized based on the stake during the previous epoch

This maybe an acceptable/fair behavior if applying the fix makes the code complex. discussion needed.

### LINKED ISSUE #178: Tx is reverted because rewards > reward pool

### LINKED ISSUE #207: Reward pool is reduced every reveal
**Describe the bug**
Reward pool is reduced every reveal

**To Reproduce**
1. First staker reveals
2. First staker rewards come from reward pool
3. second staker reveals
4. Second staker rewards come from reward pool = reward pool - first staker reward

**Expected behavior**
Need to find a fair solution.

### LINKED ISSUE #210: StakeGettingReward is not in-line when staker is penalized for some assets and rewarded for other
**Describe the bug**
Consider we have 5 assets 
Staker vote correct for 4, and wrong for 1
Then as per our current code, we don't update stake getting reward for those 4 correct votes in givePenalties but in giveRewards, we do give rewards for those 4 correct votes.

GivePenalties
https://github.com/razor-network/contracts/blob/115ad2185d7b09ad10a830f52c74605e63dd49b1/contracts/Core/RewardManager.sol#L296

GiveRewards
https://github.com/razor-network/contracts/blob/115ad2185d7b09ad10a830f52c74605e63dd49b1/contracts/Core/RewardManager.sol#L119

This breaks the following formula 
https://github.com/razor-network/contracts/blob/115ad2185d7b09ad10a830f52c74605e63dd49b1/contracts/Core/RewardManager.sol#L159

Which is written considering
Summation of thisStaker.stake for all stakers eligible for reward = StakeGettingReward

### LINKED ISSUE #211: Rewards are calculated at current stake of epoch for votes of last epoch.
As per our current code, while calculating reward, we consider the current stake of staker in that epoch for votes of last.
Ideally, we should consider the stake of the last epoch.

Rewards
https://github.com/razor-network/contracts/blob/115ad2185d7b09ad10a830f52c74605e63dd49b1/contracts/Core/RewardManager.sol#L159

How this can be attacked is
To get more rewards for my votes from the last epoch, I will just increase my stake for this epoch and get a piece of someones else from rewardpool.

We do have the same issue with penalties as well.
https://github.com/razor-network/contracts/issues/168

### LINKED ISSUE #213: [Proposal] Change penalty mechanism to influence based
There is a fundamental flaw in the penalty/reward mechanism, which cannot be fixed easily.
Currently, The median is used to penalize stakers.
However, the highest staker has a huge influence over deciding median. Hence they rarely get penalized and always get rewarded.
Moreover, they have a higher chance of getting block rewards.

This is causing a centralization issue. The stake of the highest staker goes on increasing over time. Once it reaches 50% of the total stake, the protocol breaks down.

The simulation provided here: https://github.com/razor-network/simulations/blob/master/rewards.ipynb

The original mechanism to be used is described in whitepaper v0.4 https://github.com/razor-network/whitepaper/blob/master/Razor%20Network%20Whitepaper%200.4.pdf section 2.4.3

With some changes:

- formula for age, maturity(aka reputation) are modified. please check simulation
- block rewards will continue to be calculated on stake only. This is to cut off the feedback loop which is causing centralization issues in the current mechanism.

The simulation here: https://github.com/razor-network/simulations/blob/master/influence2.ipynb

Open to feedback and alternative approach

hrishikeshio:
Fixes #213

Also Fixes #168 
Fixes #178 
Fixes  #207
Fixes #210 
Fixes  #211
Implement an influence/reputation-based mechanism.
Tests are passing. But some tests may require further modifications to cover more cases.


contracts/Core/StakeManager.sol
```diff
@@ -20,7 +20,18 @@ contract StakeManager is Initializable, ACL, StakeStorage {
     IRewardManager public rewardManager;
     RAZOR public razor;
     IVoteManager public voteManager;
-
+    //[math.floor(getMaturity2(i*10000)) for i in range(1,100)]
+    uint256[] public maturities =
```
SamAg19, line 24: How are we using maturities in the new mechanism? And any reason to select these numbers?

hrishikeshio: Current we are using sqrt(). Since the number of input values are limited, we are using a lookup table instead of calculating sqrt, which will require external library.

contracts/Core/RewardManager.sol
```diff
@@ -282,75 +158,50 @@ contract RewardManager is Initializable, ACL, RewardStorage {
...
-                uint256 prevStakeGettingReward = stakeGettingReward;
-                stakeGettingReward = stakeGettingReward + (previousStake); //*(1 - y);
-                emit StakeGettingRewardChange(
-                    epoch,
-                    prevStakeGettingReward,
-                    stakeGettingReward,
-                    block.timestamp
-                );
-            }
+            uint256 newAge = (previousAge + 10000 - (penalty));
```
SkandaBhat, line 194: What is 10000 here?

hrishikeshio: age is increased by 10,000 every epoch. This is to have enough precision since we don't have decimals in solidity.

contracts/Core/Parameters.sol
```diff
@@ -23,21 +24,22 @@ contract Parameters is IParameters, ACL {
     uint256 public override aggregationRange = 3;
     uint256 public override withdrawReleasePeriod = 5;
     uint256 public override resetLockPenalty = 1;
+    uint256 public override maxAge = 100 * 10000;
```
SkandaBhat, line 27: Why 100 * 10000?

hrishikeshio: Just for readability. 100 epochs. 10000 per epoch.

SkandaBhat:
Looks good to me. @SamAg19 @abhishekvispute please add your comments here.

contracts/Core/BlockManager.sol
```diff
@@ -197,15 +172,9 @@ contract BlockManager is Initializable, ACL, BlockStorage {
             lastVisited = sorted[i];
             accWeight = accWeight + (voteManager.getVoteWeight(epoch, assetId, sorted[i]));
```
abhishekvispute, line 173: Hey
When Staker proposes block, the medians they propose should be weighted on basis of influence, right?
If yes 
We have to change vote properties, like VoteWeights, right now it stores stake, we need influence there

contracts/Core/StakeManager.sol
```diff
@@ -436,6 +468,23 @@ contract StakeManager is Initializable, ACL, StakeStorage {
         return(numStakers);
     }
 
+    /// @return age of staker
+    function getAge(uint256 stakerId) external view returns(uint256) {
+        return stakers[stakerId].age;
+    }
+
+    /// @return maturity of staker
+    function getMaturity(uint256 stakerId) external view returns(uint256) {
```
abhishekvispute, line 477: Can you please explain
Where Maturity is being used?

As per whitepaper
maturity = f(age)
reputation = f(maturity)
influence = reputation * stake

Reputation is Penalised

In current code
I see only age and influence
Age is penalized 
Influence = Age * Stake 

Please correct me If I am wrong

hrishikeshio: Influence is used as weight.
Influence = stake * maturity

abhishekvispute: oh then why in influence we have age in place of maturity ?
line no. 485

contracts/Core/StakeManager.sol
```diff
@@ -436,6 +468,23 @@ contract StakeManager is Initializable, ACL, StakeStorage {
...
+        uint256 index = stakers[stakerId].age/10000;
+
+        return maturities[index];
+    }
+
+    /// @return influence of staker
+    function getInfluence(uint256 stakerId)  external view returns(uint256) {
+        return stakers[stakerId].age * stakers[stakerId].stake;
```
abhishekvispute, line 485: here

contracts/Core/BlockManager.sol
hrishikeshio: Correct. Fixed in https://github.com/razor-network/contracts/pull/218/commits/f8ccf9ce00ecce17230bed1d2a9199464358e940

contracts/Core/StakeManager.sol
hrishikeshio: yes. fixed in https://github.com/razor-network/contracts/pull/218/commits/f8ccf9ce00ecce17230bed1d2a9199464358e940

contracts/Core/StakeManager.sol
hrishikeshio: yes fixed in https://github.com/razor-network/contracts/pull/218/commits/f8ccf9ce00ecce17230bed1d2a9199464358e940

contracts/Core/BlockManager.sol
```diff
@@ -76,34 +72,19 @@ contract BlockManager is Initializable, ACL, BlockStorage {
         return(_blockMedians);
     }
 
-    function getLowerCutoffs(uint256 epoch) external view returns(uint256[] memory _lowerCutoffs) {
```
abhishekvispute, line 79: Hey, We also have to remove these from IBlockManager, In PR https://github.com/razor-network/contracts/pull/206 it is removed as well, Either you remove it in this PR or me at end, we just have to make sure at last it goes correct

contracts/Core/storage/RewardStorage.sol
```diff
@@ -5,6 +5,4 @@ import "../../lib/Structs.sol";
 
 contract RewardStorage {
     uint256 public blockReward;
```
abhishekvispute, line 7: Right now for storing only single param, is it worth to have separate storage contract, Can we move it to RewardManager directly as global variable ?

abhishekvispute: This doesn't impact anything but
As its inheritance, at last, it remains same

abhishekvispute: I don't remember why we are keeping it in RewardManager, 
Its ideal place is in parameters.

abhishekvispute (approved):

contracts/Core/storage/RewardStorage.sol
SkandaBhat: @abhishekvispute could you please create an issue for this?

contracts/Core/BlockManager.sol
SkandaBhat: @abhishekvispute you can cleanup in #206 .

SkandaBhat (dismissed):

contracts/Core/storage/RewardStorage.sol
hrishikeshio: > I don't remember why we are keeping it in RewardManager,
> Its ideal place would be in parameters.

yes I noticed as well. it needs to be in parameters

Issue #208, opened 2021-07-14
add escape hatch


hrishikeshio:
**Is your feature request related to a problem? Please describe.**
in case of a bug, allow admin to remove all staked tokens as a precaution. 

**Describe the solution you'd like**
during the beta phase, allow admin to withdraw all tokens. the feature should be disabled when the network is stable. Alternatively, the feature can be kept and admin access can be handed over to multisig.

adi44:
@hrishikeshio Basically we need to include a exit function for admin so in case of bug, admin can withdraw all the funds?

hrishikeshio:
> @hrishikeshio Basically we need to include a exit function for admin so in case of bug, admin can withdraw all the funds from stakemanager contract?

yes

PR #222, opened 2021-07-22
add escape hatch and tests

hrishikeshio:
solves #208

### DIFF
--- a/contracts/Core/Parameters.sol
+++ b/contracts/Core/Parameters.sol
@@ -4,6 +4,7 @@ pragma solidity ^0.8.0;
 import "./interface/IParameters.sol";
 import "./ACL.sol";
 
+
 contract Parameters is IParameters, ACL {
 
     // constant type can be readjusted to some smaller type than uint256 for saving gas (storage variable packing).
@@ -23,6 +24,7 @@ contract Parameters is IParameters, ACL {
     uint256 public override aggregationRange = 3;
     uint256 public override withdrawReleasePeriod = 5;
     uint256 public override resetLockPenalty = 1;
+    bool public override escapeHatchEnabled = true;
 
     uint32 constant private _COMMIT = 0;
     uint32 constant private _REVEAL = 1;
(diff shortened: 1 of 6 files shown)


contracts/Core/StakeManager.sol
```diff
@@ -312,6 +312,19 @@ contract StakeManager is Initializable, ACL, StakeStorage {
         emit Withdrew(epoch, stakerId, rAmount, staker.stake, block.timestamp);
     }
 
+    /// @notice remove all funds in case of emergency
```
SkandaBhat, line 315: Probably worth it to add more info as comments here. 

"Remove all funds in case of emergency, to safeguard users funds in case of a bug"

SkandaBhat (approved):
LGTM

contracts/Core/Parameters.sol
```diff
@@ -47,62 +49,66 @@ contract Parameters is IParameters, ACL {
...
+    function setGracePeriod(uint256 _gracePeriod) external onlyRole(DEFAULT_ADMIN_ROLE) {
         gracePeriod = _gracePeriod;
     }
 
     function setAggregationRange(uint256 _aggregationRange) external onlyRole(DEFAULT_ADMIN_ROLE) {
         aggregationRange = _aggregationRange;
     }
 
+    function disableEscapeHatch() external onlyRole(DEFAULT_ADMIN_ROLE) {
+        escapeHatchEnabled = false;
```
SamAg19, line 109: A toggle like feature can work here, right? Because once this is disabled, it can not be enabled. Or is this what we actually want to happen?

SkandaBhat: Yes, this is expected imo

hrishikeshio: It's what's supposed to happen. We will disable the escape hatch once the network is stable and without issues.

SamAg19: Alright

SamAg19 (approved):

SamAg19:
LGTM

Issue #174, opened 2021-06-24
Make Smart Contracts Pausable.


adi44:
since we are using initialize function, i think we should also introduce a mechanism to pause smart contracts. it will help us to update addresses of contracts if needed seamlessly. For example, let us say we want to update the address of stake manager in other contract, there might be possibility that someone can make transaction on stake manager during the duration we initiate the transaction for updating the address which might lead to loss of funds of the user. so in order to prevent that pausable smart contracts would be useful. we can do this once we go for mainnet launch.

adi44:
Add modifier to all the external and statechanging functions.

adi44:
https://docs.openzeppelin.com/contracts/2.x/api/lifecycle

PR #227, opened 2021-07-23
add pause.sol and fix lint

hrishikeshio:
Solves #174


SkandaBhat:
We need more tests to cover pausable contracts.

contracts/Core/StakeManager.sol
```diff
@@ -392,7 +395,8 @@ contract StakeManager is Initializable, ACL, StakeStorage {
         uint256 _stake,
         string memory _reason,
         uint256 _epoch
-    ) external onlyRole(parameters.getStakeModifierHash()) {
+    ) external onlyRole(parameters.getStakeModifierHash())
```
abhishekvispute, line 398: Though its accessible to RewardManager only, We should pause this as well right ? WhenNotPaused() maybe wont work directly here as we want other contract functions to go on, I think we should have one if inside this block . like ``` if paused => dont change stake```

SamAg19: I think this can fixed with the new influence mechanism where in the influence is changing, not the stake unless they add stake (which will be paused) or receive block reward.

contracts/Core/BlockManager.sol
```diff
@@ -10,10 +10,11 @@ import "./storage/BlockStorage.sol";
 import "../lib/Random.sol";
 import "../Initializable.sol";
 import "./ACL.sol";
+import "../Pause.sol";
 
 
-contract BlockManager is Initializable, ACL, BlockStorage {
-    
+contract BlockManager is Initializable, ACL, BlockStorage, Pause {
```
SamAg19, line 16: Why are we inheriting Pause.sol when we are not pausing any functions in blockManager

contracts/Core/RewardManager.sol
```diff
@@ -8,12 +8,13 @@ import "./interface/IVoteManager.sol";
...
 import "../Initializable.sol";
 import "./ACL.sol";
+import "../Pause.sol";
 
 /// @title StakeManager
 /// @notice StakeManager handles stake, unstake, withdraw, reward, functions
 /// for stakers
 
-contract RewardManager is Initializable, ACL, RewardStorage {
+contract RewardManager is Initializable, ACL, RewardStorage, Pause {
```
SamAg19, line 17: Same question goes for RewardManager

contracts/Core/BlockManager.sol
hrishikeshio: removed

contracts/Core/RewardManager.sol
hrishikeshio: removed

contracts/Core/StakeManager.sol
hrishikeshio: its fine. even if attacker manages to set their stake very high they will not be able to withdraw funds since withdraw/exit functions will be paused.

abhishekvispute: hmm yes, he won't be able to withdraw 
but maybe there is a small escape here
thing is we kind of have liquid staking due to sRZR,
so I set my stake high, the value of my sRZR goes up.
and then I am free to exchange them.

abhishekvispute (dismissed):

SamAg19 (approved):

contracts/Core/StakeManager.sol
hrishikeshio: Does srzr have erc20pausabl

abhishekvispute: Not right now 
It inherits normal ERC20 now 
We can add ERC20pausable extension.

https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/token/ERC20/extensions/ERC20Pausable.sol

This will allow us complete control over a transfer .

PR #230, opened 2021-07-23
burn half the stake

### LINKED ISSUE #220: Half of the slashed stake does not get burnt
@GauravJain9 please add more info here.

**Describe the bug**

**To Reproduce**

**Expected behavior**

**Screenshots**

**Additional context**

hrishikeshio:
fix linting errors
move slash() to stakemanager since it was using lot of functions from there
remove transferBounty() as it doesnt require separate function anymore
create _setStakerStake internal function as required by slash
change appropriate permissions
fixes #220


abhishekvispute (dismissed):

contracts/Core/StakeManager.sol
```diff
@@ -393,30 +394,35 @@ contract StakeManager is Initializable, ACL, StakeStorage {
...
+        if (halfPenalty == 0) return;
+
+        _setStakerStake(id, newStake, "Slashed", epoch);
+        //reward half the amount to bounty hunter
+        //please note that since slashing is a critical part of consensus algorithm,
+        //the following transfers are not `reuquire`d. even if the transfers fail, the slashing
+        //tx should complete.
+        razor.transfer(bountyHunter, halfPenalty);
+        //burn half the amount
+        razor.transfer(parameters.burnAddress(), halfPenalty);
```
abhishekvispute, line 425: Hey, Is there a reason 
For choosing this manual transfer for burn

Rather than using one provided by openZap
https://github.com/OpenZeppelin/openzeppelin-contracts/blob/566a774222707e424896c0c390a84dc3c13bdcb2/contracts/token/ERC20/ERC20.sol#L274

If there is, we have to change it for stakedToken as well

https://github.com/razor-network/contracts/blob/808f9289d563a3ac3c3e5306684b4963c5e3f106/contracts/Core/StakeManager.sol#L294

abhishekvispute: With this manual transfer totalSupply remains the same.
With using openZep , totalSupply is decreased by that amount.
so at least in stakedToken we want this behavior, as there we have totalSupply as deciding factor in cal.

Do we want the same for Razor as well? 
IMO we should, as the value of Razor will increase with each burn with it.

hrishikeshio: _burn() is an internal function.
It's not accessible since we did not inherit Burnable.sol for $RAZOR https://github.com/OpenZeppelin/openzeppelin-contracts/blob/566a774222707e424896c0c390a84dc3c13bdcb2/contracts/token/ERC20/extensions/ERC20Burnable.sol

abhishekvispute: Yes, ERC20 have _burn as internal function
We can have one wrapper around it and have it permission based.
Like this maybe
https://github.com/razor-network/contracts/blob/808f9289d563a3ac3c3e5306684b4963c5e3f106/contracts/StakedToken.sol#L16

hrishikeshio: Razor token is already deployed on mainnet. It doesn't have burn function.

abhishekvispute: Oh yes

SamAg19 (approved):

abhishekvispute (approved):

SamAg19 (approved):

PR #241, opened 2021-07-26
Add new state

### LINKED ISSUE #105: To ensure all jobs are being fulfilled at the time of block proposal.
**Describe the bug**
A proposed staker can propose a block that does not fulfill all the jobs if confirmed. There needs to be check regarding this before a block is proposed to the network

**To Reproduce**
1. Propose a block that fulfills only 2 jobs, even though there are 4 jobs ongoing in the network.
2. The proposal is successful.

**Expected behavior**
The staker should not be able to propose such a block as all jobs need to be present in each block proposal.

** Solution **
1. Introduce a new state (#190)
2. Keep track of active assets in each epoch.

SamAg19:
Fixes #105 

# **1. New state - 'Confirm'**

A new state called "Confirm" is being introduced where the staker, whose proposed block has successfully passed through the dispute state and considered valid, will have to call the *claimBlockReward* function which confirm the block and add the block reward to the staker's stake

Epoch Cycle

# **2. Active Assets**

With the introduction of this PR, we can now keep track of the number of active assets that the stakers need to fulfill. The number of active assets is updated at the end of the epoch in the "Confirm" State. To make sure that the number remains consistent throughout the epoch, a collection or activation/deactivation of an asset can't be done immediately. Here is the new process of creating a collection:- 

Collection Creation

Here is the process of activating/deactivating an asset:-

AssetStatus

Stakers will report values for collections only. 
Since jobs are not being reported on anymore, creation of a job can be done instantly.


test/VoteManager.js
```diff
@@ -265,9 +317,13 @@ describe('VoteManager', function () {
         const ageBefore = (await stakeManager.stakers(stakerIdAcc3)).age;
         const ageBefore2 = (await stakeManager.stakers(stakerIdAcc4)).age;
         await mineToNextState(); // dispute
+        await mineToNextState(); // confirm
+
+        await blockManager.connect(signers[3]).claimBlockReward();
+
         await mineToNextState(); // commit
         epoch = await getEpoch();
-        const votes = [100, 200, 300, 400, 500, 600, 700, 800, 900];
```
hrishikeshio, line 270: why was this modified?

test/VoteManager.js
```diff
@@ -152,24 +139,28 @@ describe('VoteManager', function () {
...
-        const staker = await stakeManager.getStaker(stakerIdAcc3);
+        let staker = await stakeManager.getStaker(stakerIdAcc3);
         const { biggestInfluencerId } = await getBiggestInfluenceAndId(stakeManager);
 
         const iteration = await getIteration(stakeManager, random, staker);
         await mineToNextState(); // propose
         await blockManager.connect(signers[3]).propose(epoch,
-          [1, 2, 3, 4, 5, 6, 7, 8, 9],
-          [100, 200, 300, 400, 500, 600, 700, 800, 900],
+          [],
```
hrishikeshio, line 148: why was this modified?

test/VoteManager.js
```diff
@@ -116,7 +115,7 @@ describe('VoteManager', function () {
 
         const stakeBefore = (await stakeManager.stakers(stakerIdAcc3)).stake;
 
-        const votes = [100, 200, 300, 400, 500, 600, 700, 800, 900];
+        const votes = [0];
```
hrishikeshio, line 118: why was this modified?

test/StakeManager.js
```diff
@@ -505,8 +505,8 @@ describe('StakeManager', function () {
         const iteration = await getIteration(stakeManager, random, staker);
 
         await blockManager.connect(signers[4]).propose(epoch,
-          [1, 2, 3, 4, 5, 6, 7, 8, 9],
-          [100, 200, 300, 400, 500, 600, 700, 800, 900],
+          [],
```
hrishikeshio, line 508: ?

test/BlockManager.js
```diff
@@ -87,23 +87,91 @@ describe('BlockManager', function () {
...
+
+      await mineToNextState();
+      const stakerIdAcc5 = await stakeManager.stakerIds(signers[5].address);
+      const staker = await stakeManager.getStaker(stakerIdAcc5);
+
+      const { biggestInfluencerId } = await getBiggestInfluenceAndId(stakeManager);
+      const iteration = await getIteration(stakeManager, random, staker);
+      await blockManager.connect(signers[5]).propose(epoch,
+        [],
+        [],
```
hrishikeshio, line 139: ?

contracts/Core/BlockManager.sol
```diff
@@ -173,27 +174,38 @@ contract BlockManager is Initializable, ACL, BlockStorage {
...
     }
 
-    function confirmBlock() external initialized onlyRole(parameters.getBlockConfirmerHash()) {
+    function claimBlockReward() public initialized checkState(parameters.confirm()){
+        uint256 epoch = parameters.getEpoch();
+        uint256 stakerId = stakeManager.getStakerId(msg.sender);
+        require(stakerId > 0, "Structs.Staker does not exist");
+        require(blocks[epoch].proposerId == 0, "Block already confirmed");
+
+        for (uint8 i = 0; i < proposedBlocks[epoch].length; i++) {
```
hrishikeshio, line 183: every proposed block is confirmed here?

contracts/Core/AssetManager.sol
```diff
@@ -432,4 +519,16 @@ contract AssetManager is ACL, AssetStorage {
     function getNumAssets() external view returns(uint256) {
         return numAssets;
     }
+
+    function getActiveAssets() external view returns(uint256) {
```
hrishikeshio, line 523: should be getNumActiveAssets() to be consistent

contracts/Core/AssetManager.sol
```diff
@@ -432,4 +519,16 @@ contract AssetManager is ACL, AssetStorage {
...
+      return activeAssets.length;
+    }
+
+    function getActiveAssetsList() external view returns(uint256[] memory) {
```
hrishikeshio, line 527: should be getActiveAssets

contracts/Core/AssetManager.sol
```diff
@@ -306,6 +282,124 @@ contract AssetManager is ACL, AssetStorage {
...
+        delete(pendingAssetDeactivation);
+    }
+
+    function activateAssets() 
+        external
+        onlyRole(parameters.getAssetConfirmerHash()) 
+    {
+        uint256 epoch = parameters.getEpoch();
+        
+        for(uint256 i = 0; i < pendingAssetActivation.length; i++){
```
hrishikeshio, line 375: what if admin creates a large number of jobs by mistake? (50+)

hrishikeshio (changes requested):

test/VoteManager.js
SamAg19: As this is the first epoch, there wont be any active assets. The require statement in propose will cause the transaction to fail if we follow the previous convention. Even if we create a collection in the first epoch, we have to go through an epoch wherein, an empty block is proposed and confirmed before that collection is activated and stakers start reporting as activation of newly created collection takes place when a block is confirmed

test/VoteManager.js
SamAg19: No assets to report in the first epoch and votes array cant be empty so went with this.

test/StakeManager.js
SamAg19: Mentioned above

test/BlockManager.js
SamAg19: Mentioned above

contracts/Core/BlockManager.sol
SamAg19: Nope, we come out of the for loop in L191. It works same as this https://github.com/razor-network/contracts/blob/034b303a4174ae846e0ed21b013b90c68bd41715/contracts/Core/BlockManager.sol#L181

SamAg19: https://github.com/SamAg19/contracts/blob/f4223bc5fe142098309de004ddaf9f8a80f792da/contracts/Core/BlockManager.sol#L181 makes sure that this function is called only once and cant be called once a block is confirmed

contracts/Core/AssetManager.sol
SamAg19: They can be deactivated then. Anyways, stakers wont be reporting for jobs anymore, only collections will reported on so it does not affect reporting also

test/VoteManager.js
hrishikeshio: then add the first epoch as a separate test in the beginning rather than modifying a bunch of tests which are testing various things. This can cause unknown bug to escape.

contracts/Core/BlockManager.sol
hrishikeshio: the if doesnt need to be that big. it can just be if (!proposedBlocks[epoch][i].valid) return;

same in https://github.com/razor-network/contracts/blob/034b303a4174ae846e0ed21b013b90c68bd41715/contracts/Core/BlockManager.sol#L181

contracts/Core/BlockManager.sol
```diff
@@ -173,27 +174,38 @@ contract BlockManager is Initializable, ACL, BlockStorage {
         }
     }
 
-    function confirmBlock() external initialized onlyRole(parameters.getBlockConfirmerHash()) {
+    function claimBlockReward() public initialized checkState(parameters.confirm()){
```
hrishikeshio, line 177: this function should be `external` right? since its only called externally?
Also if the function is called externally then the following calculation should be done offchain. no need to do it here.
the proposed block id should be parameter of this function.

contracts/Core/BlockManager.sol
```diff
@@ -173,27 +174,38 @@ contract BlockManager is Initializable, ACL, BlockStorage {
...
+        uint256 epoch = parameters.getEpoch();
+        uint256 stakerId = stakeManager.getStakerId(msg.sender);
+        require(stakerId > 0, "Structs.Staker does not exist");
+        require(blocks[epoch].proposerId == 0, "Block already confirmed");
+
+        for (uint8 i = 0; i < proposedBlocks[epoch].length; i++) {
+            if (proposedBlocks[epoch][i].valid) {
+                require(proposedBlocks[epoch][i].proposerId == stakerId, "Block can be confirmed by proposer of the block");
```
hrishikeshio, line 185: if first block isnt proposed by staker, the entire tx will fail?

contracts/Core/AssetManager.sol
```diff
@@ -165,64 +166,37 @@ contract AssetManager is ACL, AssetStorage {
...
-            epoch,
-            collections[id].active,
-            block.timestamp
-            );
-            
+            pendingAssetDeactivation.push(id);
         }
     }
 
     function fulfillAsset(
```
hrishikeshio, line 177: please write natspec of what this function does.
Why is it required if we don't allow repeat? then no collection can be fulfilled.
Want to avoid for loops in confirmblock and remove this function if possible.

contracts/Core/AssetManager.sol
```diff
@@ -306,6 +282,124 @@ contract AssetManager is ACL, AssetStorage {
...
+                collections[numAssets].id = numAssets; 
+                collections[numAssets].name = pendingCollections[i].name; 
+                collections[numAssets].aggregationMethod = pendingCollections[i].aggregationMethod; 
+                collections[numAssets].epoch = currentEpoch;
+                collections[numAssets].creator = pendingCollections[i].creator;
+                collections[numAssets].repeat = pendingCollections[i].repeat;
+                collections[numAssets].assetType = uint256(assetTypes.Collection);
+                collections[numAssets].active = true;
+                uint256[] memory jobIDs = pendingCollections[i].jobIDs;
+                for(uint256 j = 0; j < jobIDs.length; j++){
```
hrishikeshio, line 306: Code complexity O(m*n)
anything above O(n) should be avoided.

contracts/Core/AssetManager.sol
```diff
@@ -306,6 +282,124 @@ contract AssetManager is ACL, AssetStorage {
         );
     }
 
+    function addPendingCollections() 
+        external
+        onlyRole(parameters.getAssetConfirmerHash()) 
+    {
+        if(numPendingCollections != 0)
```
hrishikeshio, line 289: if(numPendingCollections == 0) return/require

contracts/Core/AssetManager.sol
```diff
@@ -306,6 +282,124 @@ contract AssetManager is ACL, AssetStorage {
...
+                id,
+                epoch,
+                jobs[id].active,
+                block.timestamp
+                );
+            }
+            else{
+                collections[id].active = false;
+
+                for(uint j = 0; j < activeAssets.length; j++){
```
hrishikeshio, line 350: avoid O(m*n)

contracts/Core/AssetManager.sol
hrishikeshio: what if its not deactivated?
to be on safer side, there should be a cap on how many jobs/collections added in single epoch. and throw if additional jobs are added

contracts/Core/BlockManager.sol
SamAg19: The first valid block, not the first block itself. The staker that has proposed the first valid block can only confirm the block

contracts/Core/AssetManager.sol
SamAg19: I am having to do it in this manner as if a collection is deactivated, it must be removed from activeAssets list and since we don't know the position of the id in the array, i am having to run a for loop to delete it from the list.

contracts/Core/AssetManager.sol
SamAg19: Been done this way to mark each jobID that is a part of the collection which is being done by a mapping.

contracts/Core/BlockManager.sol
SamAg19: We still need to verify onchain whether the block ID that is being sent is the first valid block. Otherwise, essentially anyone can confirm their block if it is in the proposal list.

abhishekvispute:
Hey @SamAg19 
Can you please update the PR description with present status, like 
we removed pendingCollections  later, there are going to be collections only, jobs are for record purposes only

abhishekvispute:
Hey @SamAg19
Can you please update the desc of PR with final solution, so we can check and finalize if it's doing as expected?

PR #253, opened 2021-07-27
Randao implementation

### LINKED ISSUE #252: Use commit secrets as a source of randomnes
**Describe the bug**
Currently block.number is used as a source of randomness which can be manipulated by miners.
We should use the secrets used by all stakers  in commit phase concatenated as a source of randomness.
This is coincidently very similar to the randao algorithm https://github.com/randao/randao
If this is unfeasible due to gas limits when there is a high number of stakers, we can potentially select secrets of 5 block proposers in previous epoch.

hrishikeshio:
fixes #252 
note that gas optimisations are not added and it uses all secrets as seed.
also asset allocation is being done using randao on previous epoch which is not secure since its already known.
randomlyassignassets should be disabled for v1.
moved blockreward to parameters
added randao penalty (stakers commits and doesnt reveal)
added check in propose to make sure staker revealed in this epoch (may try to skip commit in order to avoid penalties)


contracts/Core/VoteManager.sol
```diff
@@ -142,4 +157,13 @@ contract VoteManager is Initializable, ACL, VoteStorage {
         // epoch -> asset -> stakeWeight
         return (totalInfluenceRevealed[epoch][assetId]);
     }
+
+    function _getRandaoHash() internal view returns (bytes32) {
```
abhishekvispute, line 161: Hey @hrishikeshio 
IMO this would not work for the reveal - isAssetAllotedToStaker

As per this code, we are having secrets of this epoch as seed, which is being updated in the same state where it is being used.
It should be the secret of the previous epoch right.
Epoch 1
secrets[4] = [A,B,C,D]

Epoch 2 
Staker 1 reveals with secret E and considering seed [A,B,C,D]
secrets[4] = [E,B,C,D]

Now for the next stakers, they won't be sure on present status of seed, as it's updated on each stakers reveal, and there is no defined order for them to reveal

contracts/Core/RewardManager.sol
```diff
@@ -98,16 +90,25 @@ contract RewardManager is Initializable, ACL, RewardStorage {
...
         if (inactiveEpochs <= parameters.gracePeriod()) {
             return;
         }
         uint256 previousStake = thisStaker.stake;
+        uint256 newStake = thisStaker.stake;
         uint256 previousAge = thisStaker.age;
+        uint256 newAge = thisStaker.age;
+
+        // Not reveal penalty due to Randao
+        if (thisStaker.epochLastRevealed < thisStaker.epochLastCommitted) {
```
abhishekvispute, line 106: Hey 
Is there a scenario 
Where it would be best for staker to commit but not reveal?

hrishikeshio: Yes.
Attacker waits till everyone has revealed.
then only reveals if his reveal will make him the block proposer.

contracts/Core/VoteManager.sol
abhishekvispute: Maybe we can use stakers commitments as seed secret and commitment both have the same datatype bytes32, so theoretically it's possible.
Also this would solve another issue of knowing seed before commit as well

hrishikeshio: Correct. But we are going to disable assign assets randomly right?

abhishekvispute: If we use commitments as seed, all stakers would have to collude at once to manipulate prng

hrishikeshio: no you can again wait till everyone has committed and then choose commitment which is favourable to you.

abhishekvispute: > no you can again wait till everyone has committed and then choose commitment which is favourable to you.

Oh Yes

contracts/Core/RewardManager.sol
abhishekvispute: Oh yes

hrishikeshio:
now secrets are not stored. instead each secret is concated with previous concatenated secrets hash on reveal()

hrishikeshio:
@abhishekvispute @SkandaBhat @SamAg19 
merge conflicts resolved please approve

abhishekvispute (approved):

SkandaBhat (approved):

Issue #259, opened 2021-07-29
Dont allow zero in propose and reveal


SkandaBhat:
More info to be added here.

shekhar2807:
closing via #262

PR #262, opened 2021-07-29
test Case for if reported value is zero

### LINKED ISSUE #205: What if reported value is 0
**Describe the bug**
Currently, we use value 0 as null in many instances. What will happen if the price of an asset is genuinely 0 or reported as 0 due to malfunction at the data source?

Write tests and in case of failure, write a solution.

GaurShekhar:
fix #205 

Description:
Added test cases for 'if the reported value is zero'

P.S : As of now they are negative test cases , once we add the require in reveal that zero value can not be reported they should pass.

### DIFF
--- a/test/VoteManager.js
+++ b/test/VoteManager.js
@@ -36,9 +36,11 @@ describe('VoteManager', function () {
     let assetManager;
     let initializeContracts;
     let revealedAssetsThisEpoch = {};
+
     let blockThisEpoch = {
       ids: [], medians: [],
     };
+    const toBeDisputedAssetId = 0;
     let revealedVotesForStaker = [];
     before(async () => {
       ({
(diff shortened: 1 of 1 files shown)


test/VoteManager.js
```diff
@@ -533,6 +535,192 @@ describe('VoteManager', function () {
...
+        const proof = [];
+        for (let i = 0; i < votes.length; i++) {
+          proof.push(tree.getProofPath(i, true, true));
+        }
+
+        const assignedAssets = await getAssignedAssets(numAssets, stakerIdAcc7, votes, proof, maxAssetsPerStaker, random);
+        const assigneedAssetsVotes = assignedAssets[0];
+        const assigneedAssetsProofs = assignedAssets[1];
+
+        const tx = voteManager.connect(signers[7]).reveal(epoch, tree.root(), assigneedAssetsVotes, assigneedAssetsProofs,
```
abhishekvispute, line 585: Hey @GaurShekhar  
Can you point me to a line in the contract 
Which is reverting this?

abhishekvispute: oh okay so the given require is yet to be added 
We are waiting on that change to be made first ?

test/VoteManager.js
```diff
@@ -533,6 +535,192 @@ describe('VoteManager', function () {
...
+        const epoch = await getEpoch();
+
+        const sortedVotes = [toBigNumber(toBeDisputedAssetId * 0)];
+
+        const tx = blockManager.connect(signers[8]).giveSorted(epoch, toBeDisputedAssetId, sortedVotes);
+
+        await assertRevert(tx, 'sorted[i] is not greater than lastVisited');
+      });
+
+      it('if the reported value is zero , nothing should work in next epoch', async function () {
```
hrishikeshio, line 638: why nothing should work in next epoch?

GaurShekhar: Will rename the testCase name to "if reported value is zero , only commit should work in next epoch " 

Does this sound good ?

GaurShekhar: Will rename the testCase name to "if reported value is zero , only commit should work in next epoch " 

Does this sound good ?

hrishikeshio: I dont understand this testcase? You are writing a test case that should pass because of a bug existing in the code.

test/VoteManager.js
GaurShekhar: Yes, this was being discussed in the meeting to push the negative test cases for this issue and then we will make the required changes in contract part .

test/VoteManager.js
GaurShekhar: No, I am only checking that only commit should work in each epoch if the reported value is zero but as of now we don't have any require in reveal for the reported value .


In both these testCases I have written positive test cases for commit and negative testCases for remaining of the cycle (reveal, propose, dispute ).

hrishikeshio: please add require() statements too to fix the bug

hrishikeshio: #259

GaurShekhar: Sure.

SkandaBhat: > I dont understand this testcase? You are writing a test case that should pass because of a bug existing in the code.

@GaurShekhar "Nothing should work in next epoch" is a little bit vague. Please write a description that does not have ambiguity.

GaurShekhar:
fix #259

abhishekvispute:
Lets merge #253 first, and later this 
so these tests would pass for propose

test/VoteManager.js
```diff
@@ -652,6 +652,146 @@ describe('VoteManager', function () {
         assertBNEqual(balanceAfterAcc10, balanceBeforeAcc10.add(slashPenaltyAmount.div('2')),
           'the bounty hunter should receive half of the slashPenaltyAmount of account 4');
       });
+
+      it('if reported value is zero,should be able to commit', async function () {
```
hrishikeshio, line 656: remove this testcase. there is no "reported value" input in commit

test/VoteManager.js
```diff
@@ -652,6 +652,146 @@ describe('VoteManager', function () {
...
+          proof.push(tree.getProofPath(i, true, true));
+        }
+
+        const tx = voteManager.connect(signers[8]).reveal(epoch, tree.root(), votes, proof,
+          '0x727d5c9e6d18ed15ce7ac8d3cce6ec8a0e9c02481415c0823ea49d847ccb9ddd',
+          signers[8].address);
+
+        await assertRevert(tx, 'revert');
+      });
+      it('if the reported value is zero, should not be able to propose', async function () {
```
hrishikeshio, line 701: What is reported value? explicitly mention proposed value

test/VoteManager.js
```diff
@@ -652,6 +652,146 @@ describe('VoteManager', function () {
...
+        const tx = blockManager.connect(signers[8]).propose(epoch,
+          [19, 20, 21, 22, 23, 24, 25, 26, 27],
+          medians,
+          iteration,
+          biggestInfluencerId);
+
+        await assertRevert(tx, 'revert');
+      });
+
+      it('if the reported value is zero, should not be able to dispute', async function () {
```
hrishikeshio, line 721: what is reported value? do you mean value in reveal or propose?

test/VoteManager.js
```diff
@@ -652,6 +652,146 @@ describe('VoteManager', function () {
...
+        const epoch = await getEpoch();
+
+        const sortedVotes = [toBigNumber('0')];
+
+        const tx = blockManager.connect(signers[9]).giveSorted(epoch, 1, sortedVotes);
+
+        await assertRevert(tx, 'sorted[i] is not greater than lastVisited');
+      });
+
+      it('if the reported value is zero, only commit should work in next epoch', async function () {
```
hrishikeshio, line 732: whats reported value here? revealed/propsoed/something else?

test/VoteManager.js
GaurShekhar: I thought votes array is used in commitment , so i should add this as a testCase that staker can commit or not if the reported value is zero. If you say , I will remove it.

hrishikeshio: what do you mean by reported value?

test/VoteManager.js
GaurShekhar: This is the next epoch, where I check if the reported values are zero, everything should work as we expected.
In this epoch everything starts  fresh ,to check that penalties should not be applies in this epoch for staker for reporting zero value. As the staker can commit in next epoch after reporting zero value in previous epoch , I thought this was the test case we were mentioning to add here.

test/VoteManager.js
GaurShekhar: https://github.com/razor-network/contracts/pull/262/files#diff-f683c8d5b95777f9e626d97eb2b74a82ffe9f58b93fe5f4a6a704f90885120baR667

This array .

test/VoteManager.js
hrishikeshio: there is no such thing as reported value. mention what do you mean exactly.  revealed/proposed value

test/VoteManager.js
```diff
@@ -652,6 +652,143 @@ describe('VoteManager', function () {
...
+        const epoch = await getEpoch();
+
+        const sortedVotes = [toBigNumber('0')];
+
+        const tx = blockManager.connect(signers[9]).giveSorted(epoch, 1, sortedVotes);
+
+        await assertRevert(tx, 'sorted[i] is not greater than lastVisited');
+      });
+
+      it('if the revealed/proposed value is zero, only commit should work in next epoch', async function () {
```
hrishikeshio, line 729: revealed or proposed?

SkandaBhat (approved):
lgtm

hrishikeshio (approved):

PR #267, opened 2021-08-03
Added Event for delegation Accepted

### LINKED ISSUE #264: Emit an event on calling setDelegationAcceptance
As per the current contracts, no event is being emitted from contracts on `setDelegationAcceptance`.

It would be good to have an event that gets emitted while calling `setDelegationAcceptance`. This would help while getting stakerInfo.

adi44:
Fixes #264

### DIFF
--- a/contracts/Core/StakeManager.sol
+++ b/contracts/Core/StakeManager.sol
@@ -330,10 +330,9 @@ contract StakeManager is Initializable, ACL, StakeStorage, Pause {
     /// @notice Used by staker to set delegation acceptance, its set as False by default
     function setDelegationAcceptance(bool status) external {
         uint256 stakerId = stakerIds[msg.sender];
-        address stakerAddress = stakers[stakerId]._address;
         require(stakerId != 0, "staker id = 0");
         stakers[stakerId].acceptDelegation = status;
-        emit DelegationEnabled(stakerId, stakerAddress, status);
+        emit DelegationEnabled(stakerId, msg.sender, status);
     }
 
     /// @notice Used by staker to set commision for delegation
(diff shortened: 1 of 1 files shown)


contracts/Core/StakeManager.sol
```diff
@@ -328,8 +330,10 @@ contract StakeManager is Initializable, ACL, StakeStorage, Pause {
     /// @notice Used by staker to set delegation acceptance, its set as False by default
     function setDelegationAcceptance(bool status) external {
         uint256 stakerId = stakerIds[msg.sender];
+        address stakerAddress = stakers[stakerId]._address;
         require(stakerId != 0, "staker id = 0");
         stakers[stakerId].acceptDelegation = status;
+        emit DelegationEnabled(stakerId, stakerAddress, status);
```
abhishekvispute, line 336: we can directly send msg.sender as address here @adi44

contracts/Core/StakeManager.sol
```diff
@@ -328,8 +330,10 @@ contract StakeManager is Initializable, ACL, StakeStorage, Pause {
     /// @notice Used by staker to set delegation acceptance, its set as False by default
     function setDelegationAcceptance(bool status) external {
         uint256 stakerId = stakerIds[msg.sender];
+        address stakerAddress = stakers[stakerId]._address;
```
abhishekvispute, line 333: ..

adi44: yeah. we can pass msg.sender

contracts/Core/StakeManager.sol
```diff
@@ -134,6 +134,8 @@ contract StakeManager is Initializable, ACL, StakeStorage, Pause {
 
     event Delegated(uint256 epoch, uint256 indexed stakerId, address delegator, uint256 previousStake, uint256 newStake, uint256 timestamp);
 
+    event DelegationEnabled(uint256 indexed stakerId, address staker, bool delegationEnabled);
```
abhishekvispute, line 137: Given event is also going to be fired when delegation is disabled 
so maybe "DelegationEnabled" is not correct as the name 
can we have DelegationAcceptanceChanged ?

adi44:
Yeah. Even I was thinking of something like this

abhishekvispute (approved):

hrishikeshio (approved):

PR #283, opened 2021-08-05
Updated votemanager test case according to minstake

### LINKED ISSUE #282: Update VoteManager Test Cases
Since after the randao implementation the minstake value was changed from 100 to 1000 due to which some of the test cases of vote manager started failing. Need to update the test cases according to new implementation

GauravJain9:
fixes #282


test/VoteManager.js
```diff
@@ -389,6 +389,7 @@ describe('VoteManager', function () {
       it('should not be able to commit if stake is below minstake', async function () {
         await mineToNextEpoch();
         const epoch = await getEpoch();
+        await parameters.setMinStake(tokenAmount('100'));
```
hrishikeshio, line 392: Why change the parameters? Change the staked amount instead.
Adhoc changes to the parameter need to be avoided. This test is not about parameters contract.

SkandaBhat: I had suggested changing the min stake as I assumed otherwise. This is fair, we should change the staked amount itself. cc: @GauravJain9

GauravJain9: yeah

SkandaBhat (approved):

Issue #190, opened 2021-07-06
Add new state


SkandaBhat:
**Is your feature request related to a problem? Please describe.**
Related to #105.

**Describe the solution you'd like**
A new state is to be implemented to keep track of the number of active assets in the network, thereby allowing us to ensure that at the time of block proposal, all the active assets are being fulfilled in the proposed block. In this new state, the proposer of the 1st valid block will have to claim the block reward by confirming its proposed block. It is during this state where the number of active assets are also updated for the next epoch.

SkandaBhat:
This issue is part of the solution to #105. Closing this.

PR #297, opened 2021-08-10
Add new state

### LINKED ISSUE #105: To ensure all jobs are being fulfilled at the time of block proposal.

SamAg19:
Fixes #105 
Details about the new state "confirm" is in #241

Regarding the calculation of active assets, here are the following changes:- 

- Jobs can be created anytime during an epoch but collections can be created in only dispute and confirm states. 
- Changes in the activity status of jobs and collections can only be done in dispute and confirm states.
- addition/removal of jobs from collections can only be done in confirm state


contracts/Core/AssetManager.sol
```diff
@@ -64,22 +64,35 @@ contract AssetManager is ACL, AssetStorage {
...
         string calldata selector,
-        string calldata name,
-        bool repeat
+        string calldata name
     ) external onlyRole(parameters.getAssetModifierHash()) {
-        numAssets = numAssets + 1;
         uint256 epoch = parameters.getEpoch();
-        Structs.Job memory job = Structs.Job(numAssets, epoch, url, selector, name, repeat, true, msg.sender, 0, uint256(assetTypes.Job));
-        jobs[numAssets] = job;
+        require(numAssetsCreated[epoch] < parameters.maxAssetsCreatedPerEpoch(), "Asset Creation Limit Reached");
```
abhishekvispute, line 88: Hey why do we have this limit

contracts/Core/BlockManager.sol
```diff
@@ -117,8 +118,34 @@ contract BlockManager is Initializable, ACL, BlockStorage {
...
+                continue;
+            }
+            require(proposedBlocks[epoch][i].proposerId == stakerId, "Block can be confirmed by proposer of the block");
+            _confirmBlock(epoch, i);
+            rewardManager.giveBlockReward(stakerId, epoch);
+            return;
+        }
+    }
+
+    function confirmBlock(uint256 stakerId) external initialized onlyRole(parameters.getBlockConfirmerHash()) {
```
abhishekvispute, line 138: If I understand correctly, this is called only when nobody has called claimBlockReward, right ?
Can we name it different? forcedConfirmBlock, confirmBlockLastEpoch

contracts/Core/BlockManager.sol
```diff
@@ -117,8 +118,34 @@ contract BlockManager is Initializable, ACL, BlockStorage {
...
+        uint256 epoch = parameters.getEpoch();
+
+        for (uint8 i = 0; i < proposedBlocks[epoch - 1].length; i++) {
+            if (!proposedBlocks[epoch - 1][i].valid) {
```
abhishekvispute, line 142: This for loop is common in both claimBlockReward and confirmBlock with only one diff : epoch
can we move it out with epoch as param

package-lock.json
```diff
@@ -1613,14 +1613,14 @@
       "dev": true
     },
     "node_modules/@truffle/provider": {
-      "version": "0.2.34",
-      "resolved": "https://registry.npmjs.org/@truffle/provider/-/provider-0.2.34.tgz",
-      "integrity": "sha512-078SPxa6tiRsjxGObhE79Yw26+JNVhub23AArviBPcc5EGkRzDj4Wj5NNKsZIzhK7eFy5deQkc5HtQIAnZngrQ==",
+      "version": "0.2.35",
```
abhishekvispute, line 1616: Please check if this if package-lock.json update is necessary 
I am not sure as there is no package.json update

SamAg19: This was one of the conflicts i was facing when updating the branch

contracts/Core/AssetManager.sol
```diff
@@ -64,22 +64,35 @@ contract AssetManager is ACL, AssetStorage {
...
+    }
+
+    modifier checkStates(uint256 state1, uint256 state2) {
+        require(state1 == parameters.getState() || state2 == parameters.getState(), "incorrect states");
+        _;
+    }
+
+    constructor(address parametersAddress, address blockManagerAddress) {
         parameters = IParameters(parametersAddress);
+        blockManager = IBlockManager(blockManagerAddress);
```
abhishekvispute, line 79: Where is the use of blockmanager here
I did not find any call
Please remove if not needed

contracts/Core/AssetManager.sol
SamAg19: This update has been made on the basis of this discussion
https://github.com/razor-network/contracts/pull/241#discussion_r677310860

contracts/Core/BlockManager.sol
SamAg19: Yep, makes sense

contracts/Core/AssetManager.sol
abhishekvispute: If its required we dont have to store numAssetsCreated every epoch IMO
we can define it as global variable 
and reset it to 0 in confirm

SamAg19: Noted

SamAg19: Okay the issue i thought of doing this way was if no one confirms the block, then the reset wont take place until someone commits

contracts/Core/AssetManager.sol
```diff
@@ -64,22 +64,35 @@ contract AssetManager is ACL, AssetStorage {
...
+        numAssets = numAssets + 1;
+        numAssetsCreated[epoch]++;
 
-        emit JobCreated(numAssets, epoch, url, selector, name, repeat, msg.sender, block.timestamp, assetTypes.Job);
+        Structs.Job memory job = Structs.Job(numAssets, epoch, url, selector, name, true, msg.sender, uint256(assetTypes.Job));
+        jobs[numAssets] = job;
+        emit JobCreated(numAssets, epoch, url, selector, name, msg.sender, block.timestamp, assetTypes.Job);
     }
 
     function updateJob(
```
abhishekvispute, line 98: I think this also we should restrict to dispute and confirm only like we have for activate/deactivate

SamAg19: Why exactly? Because creation of jobs does not affect active assets as no staker would report for a job anymore.

contracts/Core/BlockManager.sol
SamAg19: Anyways i dont think so this for loop will be required when hrishi's PR get merged so this should not be an issue right now

abhishekvispute:
Hey @SamAg19 
Please confirm why we did so
> Changes in the activity status of jobs and collections can only be done in dispute and confirm states.

Commit - Cant allow,  as staker won't know the present status
Reveal - Propose -> All active assets should be revealed

> Addition/Removal of jobs from collections can only be done in confirm state

Commit - Cant allow,  as staker wont know the present status
Why not reveal, and propose?

contracts/Core/AssetManager.sol
abhishekvispute: Lets say we allow update in commit as well.

Job : URL : xxx
Half of commit phase has passed
Job has been updated now 
Job : URL : yyy
Some stakers commit as per this URL

at the end the minority would be penalised for no valid reason

contracts/Core/AssetManager.sol
abhishekvispute: Hmm yes, maybe a solution we can find
First, let's see if it's needed, I am not sure if we should carry this overhead
Let's discuss it in the next call

contracts/Core/BlockManager.sol
abhishekvispute: okay cool

contracts/Core/AssetManager.sol
SamAg19: Okay yes, makes sense. I thought you were talking about createJob. Alright, noted

SamAg19:
@abhishekvispute 
> > Changes in the activity status of jobs and collections can only be done in dispute and confirm states.
> 
> Commit - Cant allow, as staker won't know the present status
> Reveal - Propose -> All active assets should be revealed

Yes, this is correct


> > Addition/Removal of jobs from collections can only be done in confirm state
> 
> Commit - Cant allow, as staker wont know the present status
> Why not reveal, and propose?

If allow this in reveal and propose, it will affect dispute state as, since the addition/removal of jobs will affect aggregation calculations, anyone can dispute the values which were based on the previous calculation

abhishekvispute:
> If allow this in the reveal and propose, it will affect dispute state as, since the addition/removal of jobs will affect aggregation calculations, anyone can dispute the values which were based on the previous calculation

How ?
Our dispute is raised on basis of votes cast in this epoch.
We give sorted votes and if the mean is not matching we consider it as a malicious proposal.
Even if we add or remove a job from collection, votes revealed won't change, 
it's constant

SkandaBhat:
Closing this. Please create a PR that solves #190. We will merge that and then solve #105

PR #300, opened 2021-08-10
Fixes inactivity penatlies bug

### LINKED ISSUE #284: staker can avoid inactivity penatly
**Describe the bug**

**Expected behavior**

SamAg19:
fixes #284


SkandaBhat:
Closing this for now since it is incomplete.

PR #305, opened 2021-08-11
Wrote testcase when no one votes

adi44:
wrote testcases when no one votes


adi44:
wrong PR created. Closing this

Issue #265, opened 2021-08-02
Gas optimisation


hrishikeshio:
**Describe the bug**
Gas usage is too high
**Expected behavior**
simplify algorithm and reduce storage usage to reduce gas cost

PR #307, opened 2021-08-11
wrote testcases when no one votes, stake is less than minStake, staker inactive for epochs more than grace period

### LINKED ISSUE #173: Write testcases  where nobody votes, or there is too low stake (1-4) and penalizeEpochs
Some additional test cases needs to be written, to check what happens when nobody votes and stake is too low

adi44:
Wrote test cases when no one votes.
When no one votes finalizeDispute transaction gets reverted and in the next epoch every thing seems working fine. Also wrote a test case for penalize Epochs where staker does not vote for more than 8 epochs which is the grace period.
Wrote test case when stake is below min stake
wrote test case when staker is inactive for epochs more than grace period

fixes #173


test/StakeManager.js
```diff
@@ -94,6 +94,17 @@ describe('StakeManager', function () {
       assertRevert(tx1, 'AccessControl');
     });
 
+    it('should not be able to stake if stake is less than min stake', async function () {
+      const epoch = await getEpoch();
+      const stakes = ['0', '1', '999'];
```
abhishekvispute, line 99: Hey 
Is it necessary to run this loop
as 0,1,999 are always less than 1000
0,1 is not a boundary case

abhishekvispute:
All else looks good,
The tests would change little once #265 is merged.
So lets wait till then

test/StakeManager.js
adi44: Will make changes

SkandaBhat (dismissed):

test/VoteManager.js
```diff
@@ -638,6 +638,84 @@ describe('VoteManager', function () {
...
+      it('No Finalise Dispute should happen if no block is proposed or no one votes', async function () {
+        const epoch = await getEpoch();
+        await mineToNextState();
+        // dispute state
+        const sortedVotes = [];
+        const tx1 = blockManager.connect(signers[3]).giveSorted(epoch, 1, sortedVotes);
+        // const firstProposedBlockIndex = await blockManager.proposedBlocks(epoch, 0);
+        // const tx2 = blockManager.connect(signers[19]).finalizeDispute(epoch, firstProposedBlockIndex);
+        assert(tx1, 'should be able to give sorted votes');
+        // assertRevert(tx2, 'Error: Transaction reverted without a reason string');
```
abhishekvispute, line 669: hey 
can we revert finalize dispute here as well

adi44: @abhishekvispute hey, I did that before, but assertRevert is not catching the error, since there is no reason string returned. I even tried with the Error that came on revert but still it wasn't able to catch that. so I had to comment it.

abhishekvispute: hmm don't give reason then 
only assetRevert

adi44: I tried that also assertRevert(tx,'')

abhishekvispute: hmm, no don't pass reason only, check if only assertRevert(tx) is working

Check this code 
https://github.com/razor-network/contracts/blob/0f54d6b86c36c0efb30f5682843f74008533bdd8/test/helpers/testHelpers.js#L76

see if there is no reason, it not checked, so mostly it should work

adi44: ok

abhishekvispute (approved):

SkandaBhat (approved):

Issue #250, opened 2021-07-27
Fix security issues reported by mythx


hrishikeshio:
build is currently failing due to security issues reported by mythx

SkandaBhat:
https://swcregistry.io/docs/SWC-123

adi44:
@SkandaBhat Can I take up this issue?

hrishikeshio:
> @SkandaBhat Can I take up this issue?

yes

adi44:
SWC-123 warnings say that the require statement should be in the current function where it needed to be checked. And the require statement should be true always, which means it should not be violated.
Coming to warnings that says that we should add the visibility of constructor, from solidity versions after 0.7.0, the visibility for constructor plays no role. Also since the constructor is called during the deployment only, it does not need to have any visibility.

hrishikeshio:
@adi44 any update?

adi44:
> @adi44 any update?

Yes, currently there are two warnings that is related to swc-123. It says that requirement violation can be there. So as per the link the cause of this warning can be either wrong input from external contract that does not statisfy require statement and other is too strong condition. I think either we can check require in everyfunction instead of modifier or else keep as it is
https://githubmemory.com/repo/SmartContractSecurity/SWC-registry/issues/119

hrishikeshio:
ignore it

SkandaBhat:
Closing this one as Mythx isnt working out for us.

PR #308, opened 2021-08-11
Mythx warnings solved

adi44:
https://github.com/razor-network/contracts/issues/250#issuecomment-895178400


hrishikeshio:
we wont be using mythx

PR #318, opened 2021-08-13
Wrote testcases : staker should not propose when staker not elected or not revealed

### LINKED ISSUE #317: Include testcases when staker tries to propose a block, when it dont reveal and another testcase when block proposer is not elected
Include testcases when staker tries to propose a block, when it dont reveal and another testcase when block proposer is not elected

adi44:
Block Manager should not be able to propose when not elected.
Block Manager should not be able to propose without revealing.
fixes #317


test/BlockManager.js
```diff
@@ -516,5 +516,51 @@ describe('BlockManager', function () {
...
+      );
+      await voteManager.connect(signers[8]).commit(epoch, commitment1);
+      await mineToNextState();
+      await voteManager.connect(signers[8]).reveal(epoch, votes,
+        '0x727d5c9e6d18ed15ce7ac8d3cce6ec8a0e9c02481415c0823ea49d847ccb9ddd');
+      await mineToNextState();
+
+      const tx = blockManager.connect(signers[8]).propose(epoch,
+        [100, 200, 300, 400, 500, 600, 700, 800, 900],
+        iteration,
```
abhishekvispute, line 540: hey 
how it's assured that staker won't be elected proposer here?

adi44: it is because of 
if ((rand1 + 1) != stakerId) {
            return false;
        }

abhishekvispute: oh cool so you are getting iterations in the previous epoch
so seed is secret of the last commit 
hence in the new epoch, previous interaction won't be true 
right?

adi44:
Yes

test/BlockManager.js
adi44: yes

abhishekvispute: then there can be an issue 
there is a chance that the previous iteration would hold true for this epoch as well
failing this test case needlessly 
what could be a better approach 
is finding a invalid iteration in this epoch only

we can do so by adding the clone method of getIteration as getFirstInvalidIteration 
which would return the first iteration which won't hold the condition true.

adi44: But How it can happen, the previous iteration being true for this epoch.

abhishekvispute: because, for diff seeds, there can be the same iteration.
For example,
Your seed for this epoch is X and you find an iteration "i"
Then its not given that in the next epochs even if you change seed, the iteration cant be i
We find iteration from start only (0) everytime

adi44: Ohkay, if I just do, iteration - 1 in the propose transaction, that can also help. because it gives the first iteration. It can work too?

abhishekvispute: oh yes it will work for all cases except when iteration is 0

adi44: okay

abhishekvispute (dismissed):

abhishekvispute (dismissed):

test/BlockManager.js
GauravJain9: @abhishekvispute @adi44  We can create a function in utils which finds the first Wrong Iteration like "getFirstFalseIteration" by this it will be sure that staker is not elected and also in that function we need to add if statement for 0 like if getIteration returns 0 we return 1 and then a for loop may be starting from 0.
We discussed this on call once I was doing that

GauravJain9: passing iteration-1 was creating problem for 0 and I think othe than 0 also I tried that it was returning true sometimes and false sometimes so we discussed to create a function.

adi44: Other than zero, it should work as expected. For what iterations iteration - 1 was not working?

GauravJain9: It's not working for iteration = 2 also

GauravJain9: I checked now it's not working for 0 else it was working I checked 5 times. It's not working only for 0.

GauravJain9: But it's also working for 0 sometimes like 2 times out of 10

test/BlockManager.js
```diff
@@ -517,6 +517,50 @@ describe('BlockManager', function () {
...
+      await voteManager.connect(signers[8]).commit(epoch, commitment1);
+      await mineToNextState();
+      await voteManager.connect(signers[8]).reveal(epoch, votes,
+        '0x727d5c9e6d18ed15ce7ac8d3cce6ec8a0e9c02481415c0823ea49d847ccb9ddd');
+      await mineToNextState();
+      const { biggestInfluencerId } = await getBiggestInfluenceAndId(stakeManager);
+      const iteration = await getIteration(voteManager, stakeManager, staker);
+      const tx = blockManager.connect(signers[8]).propose(epoch,
+        [100, 200, 300, 400, 500, 600, 700, 800, 900],
+        iteration - 1,
```
hrishikeshio, line 539: incorrect logic. actually check whether the iteration is incorrect.

adi44: ok. Actually I wrote this testcase to cover 
require(isElectedProposer(iteration, biggestInfluencerId, proposerId), "not elected");

adi44: Yeah in that case it won't work

adi44: I will create another function with a condition

test/helpers/utils.js
```diff
@@ -105,6 +105,20 @@ const getIteration = async (voteManager, stakeManager, staker) => {
...
+const getFalseIteration = async (voteManager, stakeManager, staker) => {
+  const numStakers = await stakeManager.getNumStakers();
+  const stakerId = staker.id;
+  const influence = await stakeManager.getInfluence(stakerId);
+
+  const { biggestInfluence } = await getBiggestInfluenceAndId(stakeManager);
+  const randaoHash = await voteManager.getRandaoHash();
+  for (let i = 0; i < 10000000000; i++) {
+    const isElected = await isElectedProposer(i, biggestInfluence, influence, stakerId, numStakers, randaoHash);
+    if (isElected && i > 1) return (i - 1);
```
hrishikeshio, line 117: what if iselected= TRUE FOR i = 1
and iselected=true for i=2

adi44: I think, 
for iselected = true and i = 2, it should return 1
for iselected = true and i = 1, it should return 0
for iselected = true and i = 0, it should return 1 since negative numbers are not accepted

hrishikeshio: for iselected = true and i = 2, it should return 1

thats the problem

it will return 1. but 1 can be elected.

adi44: Oh okay

hrishikeshio (approved):

abhishekvispute (approved):

PR #324, opened 2021-08-16
Delegator TestCoverage : wrote a testcase to check if only admin can call upgradeDelegate method.

### LINKED ISSUE #323: Write a testcase, for checking if only admin can call the upgradeDelgate method in Delegator.sol
We havent covered a testcase to check if upgradeDelegate method can be called by admin only.

adi44:
In Delegator.sol, we were using the require statement to check if caller is owner or not. Here I have replaced that statement with onlyRole Modifier by ACL.sol. Also I have written a testcase for checking if only admin can call upgradeDelegate. fixes #323

### DIFF
--- a/contracts/Delegator.sol
+++ b/contracts/Delegator.sol
@@ -6,7 +6,6 @@ import "./Core/ACL.sol";
 
 contract Delegator is ACL {
     address public delegate;
-    address public owner = msg.sender;
     IAssetManager public assetManager;
 
     function upgradeDelegate(address newDelegateAddress) external onlyRole(DEFAULT_ADMIN_ROLE) {
--- a/contracts/Delegator.sol
+++ b/contracts/Delegator.sol
@@ -2,14 +2,14 @@
 pragma solidity ^0.8.0;
 
 import "./Core/interface/IAssetManager.sol";
+import "./Core/ACL.sol";
 
-contract Delegator {
+contract Delegator is ACL {
     address public delegate;
     address public owner = msg.sender;
     IAssetManager public assetManager;
 
-    function upgradeDelegate(address newDelegateAddress) external {
-        require(msg.sender == owner, "caller is not the owner");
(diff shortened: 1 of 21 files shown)


hrishikeshio (approved):

contracts/Delegator.sol
```diff
@@ -2,14 +2,14 @@
 pragma solidity ^0.8.0;
 
 import "./Core/interface/IAssetManager.sol";
+import "./Core/ACL.sol";
 
-contract Delegator {
+contract Delegator is ACL {
     address public delegate;
     address public owner = msg.sender;
```
abhishekvispute, line 9: Hey
We can remove this as well

contracts/Delegator.sol
```diff
@@ -2,14 +2,14 @@
 pragma solidity ^0.8.0;
 
 import "./Core/interface/IAssetManager.sol";
+import "./Core/ACL.sol";
```
abhishekvispute, line 5: Hey, Importing ACL is not always best approch maybe for access control. In all contracts where we have inherited ACL, a sperate copy is deployed. ACL is necessary where there are multiple roles, now here in Delegator, if its only admin, then I think ownable is better choice here, as needlessly whole acl functions, grant, revoke are deployed here

abhishekvispute: @hrishikeshio @SamAg19 what's your opinion?

hrishikeshio: better to use a single controller. Multiple controller can create bugs where different permission is given to same address across different contracts.

abhishekvispute: hmm but I think with ACL also not a single controller
multiple separate copies of ACL are deployed per contract 
As we are inheriting and not using association

ACL of Delegator won't know anything about JobConfirmer 

If we use association by creating an object from the interface then it will be a single controller

contracts/Delegator.sol
adi44: ok

hrishikeshio (approved):

abhishekvispute (approved):

PR #325, opened 2021-08-16
testcase : when staker puts wrong asset ID in givesorted or stakerID who didnt vote in current epoch.

adi44:


test/BlockManager.js
```diff
@@ -545,5 +545,69 @@ describe('BlockManager', function () {
         biggestInfluencerId);
       assertRevert(tx, 'stake below minimum stake');
     });
+    it('should not be able to give sorted votes for wrong asset ID or for stakers who didnt vote in epoch', async function () {
```
abhishekvispute, line 548: Hey 
Can divide this into 2 individual testcases.
We don't need to repeat the whole process again, in the next test state of the previous test is carried forward.

adi44: ok. Yeah Initially I didnt repeat the whole process, but then there was some issues coming up due to which I had to do it with new stakers.

abhishekvispute (approved):

SkandaBhat (approved):
lgtm

PR #327, opened 2021-08-16
Checked that contracts are initialized only once

### LINKED ISSUE #326: Write a test case to check that contracts should not be initialized again once initialized

GauravJain9:
Fixes #326 
Increases coverage on Initializable file.

### DIFF
--- a/test/VoteManager.js
+++ b/test/VoteManager.js
@@ -88,7 +88,7 @@ describe('VoteManager', function () {
       });
 
       it('should not be able to initialize contracts if they are already initialized', async function () {
-        const tx = Promise.all(await initializeContracts());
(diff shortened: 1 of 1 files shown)


test/VoteManager.js
```diff
@@ -87,6 +87,11 @@ describe('VoteManager', function () {
         await stakeManager.connect(signers[4]).stake(epoch, tokenAmount('19000'));
       });
 
+      it('should not be able to initialize contracts if they are already initialized', async function () {
+        const tx = Promise.all(await initializeContracts());
```
hrishikeshio, line 91: Promise.all() is not required here.

GauravJain9: ok so I will call initialize only on one contract (votemanager) which is already initialized @hrishikeshio

abhishekvispute (approved):

PR #334, opened 2021-08-17
Made assetId consistent | 1 to N throughout

### LINKED ISSUE #182: Inconsistent Numbering of assetId
Hi Guys,
I think
There is one issue with assetId numbering
Sometimes its starting index considered as 1, sometimes 0

In assetManager we start assetId by 1

https://github.com/razor-network/contracts/blob/6a6b116155cb0b2e889c25f6c8389efd91021943/contracts/Core/AssetManager.sol#L82

But while storing revealed votes 
We start from 0

https://github.com/razor-network/contracts/blob/6a6b116155cb0b2e889c25f6c8389efd91021943/contracts/Core/VoteManager.sol#L94

https://github.com/razor-network/contracts/blob/6a6b116155cb0b2e889c25f6c8389efd91021943/test/VoteManager.js#L129

The same can be seen in blocks, there in medians, cutoffs we assume starting index of an asset as 0.
(This only saved us while calculating penalties and rewards as there we consider vote and block storage, and as both start from 0 now, it doesn't break)

As of now, one place I found where it breaks is in dispute, in giveSorted
https://github.com/razor-network/contracts/blob/6a6b116155cb0b2e889c25f6c8389efd91021943/contracts/Core/BlockManager.sol#L176
There we ask for assetId from staker,
Here staker has to pass assetId as per 0-> NumAssets-1, but from assetManager a Staker would know it as 1->NumAssets, raising an invalid dispute
The same can be seen in tests, their we are passing assetId -1 not assetId
https://github.com/razor-network/contracts/blob/6a6b116155cb0b2e889c25f6c8389efd91021943/test/BlockManager.js#L241

Please verify if its a genuine issue, If it is we have to thoroughly check where else its breaking

abhishekvispute:
Fixes #182 

VoteManager stores votes from 0..n-1 for assets with assetids 1 to n

It mainly impacted two functions
getVoteValue 
and 
giveSortedVotes

In both places, the user earlier had to pass assetid-1 as assetid to get the required results.
both have been solved by this PR.
now user don't have to pass -1, he passes from 1 to N only, we do -1 inside
please check tests to see how its changed

### DIFF
--- a/contracts/Core/RewardManager.sol
+++ b/contracts/Core/RewardManager.sol
@@ -126,7 +126,7 @@ contract RewardManager is Initializable, ACL, Constants {
         if (mediansLastEpoch.length == 0) return;
         uint64 penalty = 0;
         for (uint8 i = 0; i < mediansLastEpoch.length; i++) {
-            uint32 voteValueLastEpoch = voteManager.getVoteValue(i, stakerId);
+            uint32 voteValueLastEpoch = voteManager.getVoteValue(i + 1, stakerId);
             // uint32 voteWeightLastEpoch = voteManager.getVoteWeight(thisStaker.id, i);
             uint32 medianLastEpoch = mediansLastEpoch[i];
             if (medianLastEpoch == 0) continue;
--- a/contracts/Core/BlockManager.sol
+++ b/contracts/Core/BlockManager.sol
@@ -91,7 +91,7 @@ contract BlockManager is Initializable, ACL, BlockStorage, StateManager {
             Structs.Vote memory vote = voteManager.getVote(lastVisitedStaker);
             require(vote.epoch == epoch, "epoch in vote doesnt match with current");
 
-            uint48 value = vote.values[assetId-1];
+            uint48 value = vote.values[assetId - 1];
             uint256 influence = voteManager.getInfluenceSnapshot(epoch, lastVisitedStaker);
             accProd = accProd + value * influence;
             accWeight = accWeight + influence;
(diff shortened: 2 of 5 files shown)


contracts/Core/RewardManager.sol
```diff
@@ -126,7 +126,7 @@ contract RewardManager is Initializable, ACL, Constants {
         if (mediansLastEpoch.length == 0) return;
         uint64 penalty = 0;
         for (uint8 i = 0; i < mediansLastEpoch.length; i++) {
-            uint32 voteValueLastEpoch = voteManager.getVoteValue(i, stakerId);
+            uint32 voteValueLastEpoch = voteManager.getVoteValue(i + 1, stakerId);
```
hrishikeshio, line 129: start the value of i from 1 instead.

hrishikeshio (changes requested):

contracts/Core/RewardManager.sol
abhishekvispute: sure we can do so, but if we do it here 
on line 131 we will have to do -1
In the end it remains the same 

so should we do it?

hrishikeshio (approved):

contracts/Core/RewardManager.sol
SamAg19, line 129: The only issue of solving it in this way is that if we decide to deactivate an asset, it will start pointing to the wrong assetID again.

SkandaBhat: Let's tackle this in a separate issue. cc: @abhishekvispute @hrishikeshio @SamAg19

SkandaBhat (approved):
lgtm

PR #335, opened 2021-08-17
Add new state

### LINKED ISSUE #190: Add new state

SamAg19:
Fixes #190


contracts/Core/StateManager.sol
```diff
@@ -29,7 +29,7 @@ contract StateManager is Constants {
     }
 
     function getState(uint32 epochLength) public view returns (State) {
-        uint8 state = uint8(((block.number) / (epochLength / NUM_STATES)));
+        uint32 state = uint32(((block.number) / (epochLength / NUM_STATES)));
```
SkandaBhat, line 32: This can be uint8. Any specific reason it was changed to uint32?

SkandaBhat (changes requested):

contracts/Core/StateManager.sol
SamAg19: With the increase in block number, the value of state also increase to the point where it surpasses 255 and starts again from 0. This change was causing the getState to return the wrong state. Essentially with every 15300 blocks that are mined, the state also resets to 0. Hence, changed it to uint32 to be safe and not to encounter such issues again

hrishikeshio: the calculation inside brackets happens at uint256. after calculation value is between (0,4). the uint8() converts it to uint8. How can it overflow?

SkandaBhat: Sameer might be correct.

Dry run:
block.number: 15300
epochLength: 300
NUM_STATES: 5

uint8 state = uint8 (15300/(300/5))
state = 255
return 255%5 = 0

Anything above this and state will start overflowing. We can perform modulus and return without storing state maybe.

hrishikeshio: Oh I see. The %5 needs to be done in the bracket then. I thought that was being done already.

uint32(((block.number) / (epochLength / NUM_STATES))%5);

SkandaBhat: Ya this works. cc @SamAg19

SamAg19: Done, implemented the change

hrishikeshio (approved):

SkandaBhat (approved):

abhishekvispute (approved):

PR #342, opened 2021-08-18
Slither

SkandaBhat:
FYI Slither also has a WIP vscode integration - https://github.com/samparsky/slither-vscode


.circleci/config.yml
```diff
@@ -84,29 +84,17 @@ jobs:
           path: coverage
           prefix: coverage
 
-  mythx:
+  eth-security-toolbox:
```
hrishikeshio, line 87: `slither` would be more appropriate name as the toolbox contains many tools

SkandaBhat: The docker image we use for this job gives us access to other tools as well. 

https://github.com/trailofbits/eth-security-toolbox

The intention was to expand this job to include one or more tools from the toolbox in the future. If we are not planning on using them, then seems reasonable to call it slither. What do you think?

hrishikeshio: this job is only running slither for now. Better to name it slither.

SkandaBhat: Sure. 👍

SkandaBhat: @hrishikeshio pls resolve and merge

hrishikeshio (approved):

Issue #350, opened 2021-08-19
[Epic] Test Coverage should be 100%


SkandaBhat:
- [x] BlockManager/ClaimBlockReward()
- [x] finalize dispute should not work if medians = 0
- [x] BlockManager/ getProposedBlock, getBlockMedians, getProposedBlockMedians[
- [x] test cases for updateCollection
- [x] StakeManager / withdraw / https://github.com/razor-network/contracts/blob/8a27a6878c3f6ebd115b8c298b725ef46b438a95/contracts/Core/StakeManager.sol#L184
- [x] StakeManager / resetLock / https://github.com/razor-network/contracts/blob/8a27a6878c3f6ebd115b8c298b725ef46b438a95/contracts/Core/StakeManager.sol#L258
- [x] StakeManager / withdraw / https://github.com/razor-network/contracts/blob/8a27a6878c3f6ebd115b8c298b725ef46b438a95/contracts/Core/StakeManager.sol#L187
- [x] check finalizeDispute should not work if total Influence Reward does not match
- [x] Initializable/ initializer (Write test case to check modifier initializer in Initializable.sol
- [x] StateManager / checkDisputeorConfirmState, notCommitState
- [x] #437
- [x] Add test with different bigInflucnerId passed in propose, preference should be given to correct ones, even if iteration for propose with incorrect bigInfluencerId is lesser
Ref : https://github.com/razor-network/contracts/issues/414#issuecomment-918263520

SkandaBhat:
@GauravJain9 what is the status on this? Why is `BlockManager/ insertAppropriately` not closed yet?

GauravJain9:
> @GauravJain9 what is the status on this? Why is `BlockManager/ insertAppropriately` not closed yet?

Actually inside this function the if block which was thought of as redundant is no more redundant and it is useful so once after testing that part (which is itself the 12th task)  both task 11 & 12 can be closed.

PR #352, opened 2021-08-20
Finalize dispute should not work if medians == 0 and getter functions should work as expected

adi44:
#350 
Wrote test case finalize dispute should not work if medians = 0
wrote testcase to check getProposedBlock, getBlockMedians, getProposedBlockMedians should work as expected

### DIFF
--- a/test/BlockManager.js
+++ b/test/BlockManager.js
@@ -2,6 +2,7 @@
 test same vote values, stakes
 test penalizeEpochs */
 
+const { assert } = require('chai');
 const {
   assertBNEqual,
   mineToNextEpoch,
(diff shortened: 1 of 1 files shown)


test/BlockManager.js
```diff
@@ -625,5 +626,44 @@ describe('BlockManager', function () {
...
+      const tx = blockManager.connect(signers[19]).finalizeDispute(epoch, 0);
+      assertRevert(tx, 'median can not be zero');
+    });
+    it('should be able to return correct data for getBlockMedians', async function () {
+      const tx = await blockManager.connect(signers[19]).getBlockMedians(await getEpoch());
+      assert(tx, 'transaction should not reverted');
+    });
+    it('getProposedBlock Function should work as expected', async function () {
+      const tx = await blockManager.connect(signers[19]).getProposedBlock(await getEpoch(), 0);
+      assert(tx, 'transaction should not get reverted');
```
hrishikeshio, line 662: doesnt make sense. check if the returned value is correct instead.

adi44: ok

Issue #343, opened 2021-08-19
In function resetlock require statement becomes unreachable 


GauravJain9:
Here, 
1st require statement :   
https://github.com/razor-network/contracts/blob/ac4e177fc55422a1c4638da7725beb58fb969105/contracts/Core/StakeManager.sol#L246

2nd require statement :
https://github.com/razor-network/contracts/blob/ac4e177fc55422a1c4638da7725beb58fb969105/contracts/Core/StakeManager.sol#L247

In 2nd require statement we are checking if staker Id is 0 or not but if staker id is 0 staker can't set a lock and it reverts from 1st require statement itself saying lock doesn't exist.
We can interchange the order of require to solve this or we can remove the 2nd reuire which saves GAS as the 2nd reuire would have already been checked when it reaches there.

PR #353, opened 2021-08-20
Increased coverage on stake manager and removed two require statements from stakeManager.sol

### LINKED ISSUE #341: In function delegate require statement becomes unreachable
Here,
1st require statement
 https://github.com/razor-network/contracts/blob/ff482d535f8491ca979b7ede067fe16241996aa5/contracts/Core/StakeManager.sol#L105
and
2nd require statement
https://github.com/razor-network/contracts/blob/ff482d535f8491ca979b7ede067fe16241996aa5/contracts/Core/StakeManager.sol#L106
The second require statement becomes unreachable because if we have a case that staker has not staked yet (i.e staker does not exist) and delegator is delegating on that staker then it will be reverted from the 1st require statement itself because staker cannot setDelegationAcceptance to true without staking (this should happen and is already tested) .

So as per discussion in testing syncup call we can interchange the order of require statements 1st to 2nd and 2nd to 1st will solve this issue.

GauravJain9:
Fixes #341 #343 
task 4,5 done for => #350 

1) Require statement as mentioned in #341 #343  (desc provided) becomes unreachable.
As per discussion in testing call we decided to remove both require statements from both functions as they are already been checked before reaching them.

2) Added test to increase coverage (task 4,5  #350 ).

### DIFF
--- a/contracts/Core/StakeManager.sol
+++ b/contracts/Core/StakeManager.sol
@@ -103,7 +103,6 @@ contract StakeManager is Initializable, ACL, StakeStorage, StateManager, Pause {
         uint256 amount
     ) external initialized checkEpochAndState(State.Commit, epoch, parameters.epochLength()) whenNotPaused {
         require(stakers[stakerId].acceptDelegation, "Delegetion not accpected");
-        require(stakers[stakerId].tokenAddress != address(0x0000000000000000000000000000000000000000), "Staker has not staked yet");
         // Step 1:  Razor Token Transfer : Amount
         require(razor.transferFrom(msg.sender, address(this), amount), "RZR token transfer failed");
 
@@ -244,7 +243,6 @@ contract StakeManager is Initializable, ACL, StakeStorage, StateManager, Pause {
     function resetLock(uint32 stakerId) external initialized whenNotPaused {
         // Lock should be expired if you want to reset
         require(locks[msg.sender][stakers[stakerId].tokenAddress].amount != 0, "Existing Lock doesnt exist");
-        require(stakers[stakerId].id != 0, "staker.id = 0");
 
         Structs.Staker storage staker = stakers[stakerId];
         IStakedToken sToken = IStakedToken(stakers[stakerId].tokenAddress);
--- a/test/StakeManager.js
+++ b/test/StakeManager.js
@@ -93,6 +93,7 @@ describe('StakeManager', function () {
       await razor.transfer(signers[7].address, stake1);
       await razor.transfer(signers[8].address, stake1);
       await razor.transfer(signers[9].address, stake1);
+      await razor.transfer(signers[12].address, stake1);
     });
 
     it('should not allow non admin to pause', async function () {
@@ -255,7 +256,7 @@ describe('StakeManager', function () {
       await assertRevert(tx, 'Existing Lock doesnt exist');
     });
 
-    it('Staker should not be able to unstake more than his stoken balance', async function () {
+    it('Staker should not be able to unstake more than his sToken balance', async function () {
       const epoch = await getEpoch();
       const stakerIdAcc1 = await stakeManager.stakerIds(signers[1].address);
       const staker = await stakeManager.getStaker(stakerIdAcc1);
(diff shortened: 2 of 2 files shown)


test/StakeManager.js
```diff
@@ -255,7 +256,7 @@ describe('StakeManager', function () {
       await assertRevert(tx, 'Existing Lock doesnt exist');
     });
 
-    it('Staker should not be able to unstake more than his stoken balance', async function () {
+    it('Staker should not be able to unstake more than his sToken balance', async function () {
```
SkandaBhat, line 259: Nomenclature: sRAZOR

test/StakeManager.js
```diff
@@ -1204,5 +1205,43 @@ describe('StakeManager', function () {
       const tx = stakeManager.connect(signers[3]).withdraw(epoch, stakerIdacc3);
       await assertRevert(tx, 'Nonpositive Stake');
     });
+
+    it('staker should not be able to withdraw more than his sToken balance', async function () {
```
SkandaBhat, line 1209: same

SkandaBhat:
> As per discussion in testing call we decided to remove both require statements from both functions as they are already been checked before reaching them.

Please provide a detailed justification as to why they are removed with code, links etc.

test/StakeManager.js
GauravJain9: > > As per discussion in testing call we decided to remove both require statements from both functions as they are already been checked before reaching them.
> 
> Please provide a detailed justification as to why they are removed with code, links etc.

Provided the detailed justification here it is also provided in issues and the link to the same was already given.

test/StakeManager.js
GauravJain9: > Nomenclature: sRAZOR
what is sRAZOR balance ? Do you mean stake of staker ? @SkandaBhat

SkandaBhat: What is sToken?

SkandaBhat (approved):

test/StakeManager.js
```diff
@@ -1204,5 +1205,43 @@ describe('StakeManager', function () {
...
+      for (let i = 0; i < WITHDRAW_LOCK_PERIOD; i++) {
+        await mineToNextEpoch();
+      }
+      epoch = await getEpoch();
+      const tx = stakeManager.connect(signers[12]).withdraw(epoch, stakerId);
+      await assertRevert(tx, 'locked amount lost');
+      await stakeManager.connect(signers[12]).resetLock(stakerId);
+    });
+
+    it('sToken burn should fail if staker calls resetLock and his sToken balance is less than sAmount', async function () {
```
abhishekvispute, line 1227: Please rename this to 
'ResetLock should fail, if delegators balance is less than the amount to be penalized'

test/StakeManager.js
abhishekvispute, line 1209: Please rename this to 
Delegators/Stakers should not be able to withdraw if their current balance is less than the locked amount.

GauravJain9: > Please rename this to
> Delegators/Stakers should not be able to withdraw if their current sRZR balance is less than the locked amount.

I am only checking for staker here so can I rename to "Stakers should not be able to withdraw if their current sRZR balance is less than the locked amount"  @abhishekvispute

abhishekvispute: yes

test/StakeManager.js
GauravJain9: Here also "ResetLock should fail, if stakers sRZR balance is less than the amount to be penalized"

abhishekvispute: cool

abhishekvispute (approved):

contracts/Core/StakeManager.sol
```diff
@@ -103,7 +103,6 @@ contract StakeManager is Initializable, ACL, StakeStorage, StateManager, Pause {
         uint256 amount
     ) external initialized checkEpochAndState(State.Commit, epoch, parameters.epochLength()) whenNotPaused {
         require(stakers[stakerId].acceptDelegation, "Delegetion not accpected");
-        require(stakers[stakerId].tokenAddress != address(0x0000000000000000000000000000000000000000), "Staker has not staked yet");
```
hrishikeshio, line 106: what if staker stakes and later withdraws his stake? would acceptDelegation  still be true?

GauravJain9: > what if staker stakes and later withdraws his stake? would acceptDelegation still be true?

yes if staker stakes and setDelegationAcceptance to true and then withdraws then still for that address delegationAcceptance will be true.

SkandaBhat (approved):

SkandaBhat:
Looks good to me.

contracts/Core/StakeManager.sol
SkandaBhat: Addressed in #366 366

PR #355, opened 2021-08-21
written test cases to check finalize dispute should not work when medians zero

adi44:
#350

### DIFF
--- a/test/BlockManager.js
+++ b/test/BlockManager.js
@@ -20,6 +20,7 @@ const {
   getEpoch,
   getBiggestInfluenceAndId,
   getIteration,
+  getFalseIteration,
   toBigNumber,
   tokenAmount,
 } = require('./helpers/utils');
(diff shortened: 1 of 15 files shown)


test/BlockManager.js
```diff
@@ -644,5 +645,47 @@ describe('BlockManager', function () {
...
+      const tx = blockManager.connect(signers[19]).finalizeDispute(epoch, 0);
+      assertRevert(tx, 'median can not be zero');
+    });
+    it('should be able to return correct data for getBlockMedians', async function () {
+      const tx = await blockManager.connect(signers[19]).getBlockMedians(await getEpoch());
+      assert(tx, [], 'transaction should return correct data');
+    });
+    it('getProposedBlock Function should work as expected', async function () {
+      const tx = await blockManager.connect(signers[19]).getProposedBlock(await getEpoch(), 0);
+      assert(tx._block.medians, [0, 0, 0, 0, 0, 0, 0, 0, 0], 'it should return correct value');
```
hrishikeshio, line 681: https://www.chaijs.com/api/assert/ assert only accepts two parameters...
Why are you introducing assert? We aren't using it anywhere right?

hrishikeshio, line 681: assert() only accepts two parameters https://www.chaijs.com/api/assert/
why are you introducing assert() are we using it in any of the tests?

adi44: We are not using it in any of the tests, I am not sure if we can check array using assertBNEqual. So I am using assert. I didn't check the number of parameters, will change it.

test/BlockManager.js
```diff
@@ -644,5 +645,45 @@ describe('BlockManager', function () {
...
+      assertRevert(tx, 'median can not be zero');
+    });
+    it('should be able to return correct data for getBlockMedians', async function () {
+      const tx = await blockManager.connect(signers[19]).getBlockMedians(await getEpoch());
+      assertDeepEqual(tx, [], 'transaction should return correct data');
+    });
+    it('getProposedBlock Function should work as expected', async function () {
+      const tx = await blockManager.connect(signers[19]).getProposedBlock(await getEpoch(), 0);
+      const medians = await blockManager.connect(signers[19]).getProposedBlockMedians(await getEpoch(), 0);
+      const { proposerId, iteration, biggestInfluence } = await blockManager.proposedBlocks(await getEpoch(), 0);
```
abhishekvispute, line 682: Hey @adi44 
Here if I understand correctly 
We are comparing output from getProposedBlock() and blockManager.proposedBlocks().
As for both data sources is common only: contract. it would be true always. 
IMO better approach would be comparing on-chain data with the local one

adi44: Yeah that can be done also, but here we are not checking if values are stored correctly. We want to check if block returns values as expected or not

abhishekvispute: oh okay I see, cool @adi44 
Should we add an assertion to check if output of getProposedBlock is same as what you calculated locally?
With that whats stored will also be verified

adi44: I think we have covered in earlier testcases

abhishekvispute: where?

abhishekvispute: Hey @adi44,
Is this solved?

adi44: yes @abhishekvispute

abhishekvispute (approved):

SamAg19 (approved):

PR #357, opened 2021-08-21
Resolved slither issue warnings 

adi44:
In this PR, I have tried to resolve Reentrancy Attack Warnings by Slither and Zero Address Check. 

Following warnings has been ignored for now.

https://app.circleci.com/pipelines/github/razor-network/contracts/431/workflows/60cc5590-fa57-479c-b40b-5439f77a429c/jobs/2475


contracts/Core/VoteManager.sol
```diff
@@ -142,7 +142,7 @@ contract VoteManager is Initializable, ACL, VoteStorage, StateManager {
         return votes[stakerId].epoch;
     }
 
-    function getRandaoHash() public view returns (bytes32) {
+    function getRandaoHash() external view returns (bytes32) {
```
SkandaBhat, line 145: I think `razor-go` needs this `getRandaoHash()` @ashish10677  Please clarify

SkandaBhat: https://github.com/razor-network/razor-go/blob/e04f1727a4eaab155852dbf34cbbc3d54ddfdf33/utils/vote.go#L43

contracts/Core/StakeManager.sol
```diff
@@ -103,21 +102,22 @@ contract StakeManager is Initializable, ACL, StakeStorage, StateManager, Pause {
         uint256 amount
     ) external initialized checkEpochAndState(State.Commit, epoch, parameters.epochLength()) whenNotPaused {
         require(stakers[stakerId].acceptDelegation, "Delegetion not accpected");
-        // Step 1:  Razor Token Transfer : Amount
-        require(razor.transferFrom(msg.sender, address(this), amount), "RZR token transfer failed");
+        require(stakers[stakerId].tokenAddress != address(0x0), "Staker has not staked yet");
+        emit Delegated(msg.sender, epoch, stakerId, stakers[stakerId].stake, block.timestamp);
```
abhishekvispute, line 106: Hey by moving this up,
It will emit old data, as the stake is being changed later
Please add it after that
And please check in other places as well if by moving anything up or down, its calculating on old data

contracts/Core/StakeManager.sol
```diff
@@ -103,21 +102,22 @@ contract StakeManager is Initializable, ACL, StakeStorage, StateManager, Pause {
...
         IStakedToken sToken = IStakedToken(stakers[stakerId].tokenAddress);
         uint256 totalSupply = sToken.totalSupply();
         uint256 toMint = _convertRZRtoSRZR(amount, stakers[stakerId].stake, totalSupply);
 
-        // Step 3: Increase given stakers stake by : Amount
+        // Step 2: Increase given stakers stake by : Amount
         stakers[stakerId].stake = stakers[stakerId].stake + (amount);
 
+        // Step 3:  Razor Token Transfer : Amount
+        require(razor.transferFrom(msg.sender, address(this), amount), "RZR token transfer failed");
```
abhishekvispute, line 117: Hey @adi44 
I am not sure if this is best impl.
We had placed this transfer before
```
stakers[stakerId].stake = stakers[stakerId].stake + (amount);
```
As stake in stakeManager should be increased only when the transfer is successful
Now with this PR, we update state first and then transferFrom happens 
I am not sure what the best practice is 
But I think if its ```subtraction``` then it should be 
1. State
2. Trasnfer

If its ```addition``` then it should be 
1. Transfer
2. State

For example : Check this code 
https://github.com/maticnetwork/contracts/blob/45aa3a62264f15b6bb4d50a3fa8f6ab7b9cde829/contracts/staking/stakeManager/StakeManager.sol#L314

Here they are doing Transfer and then State as its addition

@hrishikeshio @SkandaBhat @SamAg19
 What's your opinion ?

contracts/Core/StakeManager.sol
adi44: Ideally, Delegated should only emit the amount that has been delegated instead of whole stake. we can replace that with amount.

contracts/Core/StakeManager.sol
```diff
@@ -139,9 +139,9 @@ contract StakeManager is Initializable, ACL, StakeStorage, StateManager, Pause {
         require(sAmount > 0, "Non-Positive Amount");
         IStakedToken sToken = IStakedToken(staker.tokenAddress);
         require(sToken.balanceOf(msg.sender) >= sAmount, "Invalid Amount");
+        emit Unstaked(epoch, stakerId, sAmount, staker.stake, block.timestamp);
```
abhishekvispute, line 142: Hey 
Here either way there is no external call, so no reentrcy 
So why we are moving event up ?

contracts/Core/StakeManager.sol
```diff
@@ -183,30 +183,27 @@ contract StakeManager is Initializable, ACL, StakeStorage, StateManager, Pause {
...
             uint256 commission = (rAmount * staker.commission) / 100;
-            require(razor.transfer(staker._address, commission), "couldnt transfer");
             rAmount = rAmount - commission;
+            require(razor.transfer(staker._address, commission), "couldnt transfer");
         }
 
         //Transfer stake
         require(razor.transfer(msg.sender, rAmount), "couldnt transfer");
-
-        emit Withdrew(epoch, stakerId, rAmount, staker.stake, block.timestamp);
```
abhishekvispute, line 203: This event you have removed, but not added, please add so

contracts/Core/StakeManager.sol
```diff
@@ -183,30 +183,27 @@ contract StakeManager is Initializable, ACL, StakeStorage, StateManager, Pause {
...
 
         uint256 rAmount = _convertSRZRToRZR(lock.amount, staker.stake, sToken.totalSupply());
-        require(sToken.burn(msg.sender, lock.amount), "Token burn Failed");
+        uint256 lockAmount = lock.amount;
         staker.stake = staker.stake - rAmount;
-
         // Function to Reset the lock
         _resetLock(stakerId);
-
+        require(sToken.burn(msg.sender, lockAmount), "Token burn Failed");
```
abhishekvispute, line 190: Hey cant we place this before _resetLock(), hence we won't need to store lockAmount, locally 
_resetLock is an internal function so there is no reentrancy possibility.

contracts/Core/StakeManager.sol
adi44: Yes, we dont have any external call but there is a statechange happening for lock

contracts/Core/StakeManager.sol
adi44: yes I will add

contracts/Delegator.sol
```diff
@@ -9,6 +9,7 @@ contract Delegator is ACL {
     IAssetManager public assetManager;
 
     function upgradeDelegate(address newDelegateAddress) external onlyRole(DEFAULT_ADMIN_ROLE) {
+        require(newDelegateAddress != address(0x0), "Zero Address check");
```
abhishekvispute, line 12: Please check if there is no difference in address(0x0) and address(0), as in OZ, they are using address(0)

contracts/Core/StakeManager.sol
abhishekvispute: Hmm nope, the whole stake is only being used in RazorScan

contracts/Delegator.sol
adi44: there is no difference, we can use either 0x0 or 0

contracts/Core/StakeManager.sol
abhishekvispute: Is it given that event should be emitted before state change, is there any pattern ?
I think it's just that event should be emitted before the external call, to avoid re-entracy 
```
Reentrancy in BlockManager.claimBlockReward() (contracts/Core/BlockManager.sol#110-125):
	External calls:
	- rewardManager.giveBlockReward(stakerId,epoch) (contracts/Core/BlockManager.sol#123)
	Event emitted after the call(s):
	- BlockConfirmed(epoch,proposerId,proposedBlocks[epoch][blockId].medians,block.timestamp) (contracts/Core/BlockManager.sol#124)
```
If we start emitting events before state change, what should we emit in event, old data :)

abhishekvispute: For example 
https://github.com/OpenZeppelin/openzeppelin-contracts/blob/cec4f2ef57495d8b1742d62846da212515d99dd5/contracts/token/ERC20/ERC20.sol#L237

contracts/Core/StakeManager.sol
adi44: here they are doing addition using the internal function but we are changing in the function only.

contracts/Core/StakeManager.sol
adi44: In our case yes, because we are emitting the resultant after state change.

contracts/Core/StakeManager.sol
adi44: yes we can do that

contracts/Core/StakeManager.sol
abhishekvispute: doing addition in internal function or doing in same function, its same.

contracts/Core/StakeManager.sol
adi44: I think we should rename to updatedStake if we want whole stake. Delegated will misguide user

abhishekvispute: hmm not needed 
as in event delegated, we are maintaining it as newStake

```
 event Delegated(address delegator, uint32 epoch, uint32 indexed stakerId, uint256 newStake, uint256 timestamp);
 ```

File: contracts/Core/StakeManager.sol (hunk 5 above)
Comment (adi44) [reply contracts/Core/StakeManager.sol:203]: fixed

File: contracts/Core/StakeManager.sol (hunk 2 above)
Comment (adi44) [reply contracts/Core/StakeManager.sol]: fixed

File: contracts/Core/StakeManager.sol (hunk 5 above)
Comment (adi44) [reply contracts/Core/StakeManager.sol:203]: fixed

File: contracts/Core/StakeManager.sol (hunk 6 above)
Comment (adi44) [reply contracts/Core/StakeManager.sol]: fixed

File: contracts/Core/StakeManager.sol (hunk 4 above)
Comment (adi44) [reply contracts/Core/StakeManager.sol:142]: actually here we do have external call, parameters.withdrawLockPeriod())

**Comment by SkandaBhat** [comment]
@adi44 Please add more details in the description, around which issues are fixed in this PR. There are too many commits to get a sense of the changes.

File: contracts/Core/StakeManager.sol (hunk 3 above)
Comment (hrishikeshio) [reply contracts/Core/StakeManager.sol:117]: state changes first. interactions later.

check - effects - interactions.

doing state changes later introduces potential reentrancy vulnerability.

File: contracts/Core/StakeManager.sol [hunk 8]
```diff
@@ -103,21 +102,22 @@ contract StakeManager is Initializable, ACL, StakeStorage, StateManager, Pause {
         uint256 amount
     ) external initialized checkEpochAndState(State.Commit, epoch, parameters.epochLength()) whenNotPaused {
         require(stakers[stakerId].acceptDelegation, "Delegetion not accpected");
-        // Step 1:  Razor Token Transfer : Amount
-        require(razor.transferFrom(msg.sender, address(this), amount), "RZR token transfer failed");
+        require(stakers[stakerId].tokenAddress != address(0x0), "Staker has not staked yet");
```
abhishekvispute, line 105: this require was removed from master, please remove it from here as well, CC @GauravJain9 PR

contracts/Core/StakeManager.sol
```diff
@@ -146,10 +146,10 @@ contract StakeManager is Initializable, ACL, StakeStorage, StateManager, Pause {
         uint256 rAmount = _convertSRZRToRZR(sAmount, staker.stake, sToken.totalSupply());
         staker.stake = staker.stake - rAmount;
 
-        locks[msg.sender][staker.tokenAddress] = Structs.Lock(rAmount, epoch + (parameters.withdrawLockPeriod()));
-
         staker.epochLastUnstakedOrFirstStaked = epoch;
 
+        locks[msg.sender][staker.tokenAddress] = Structs.Lock(rAmount, epoch + (parameters.withdrawLockPeriod()));
```
abhishekvispute, line 151: Why its position is changed ?

abhishekvispute: before 
```staker.epochLastUnstakedOrFirstStaked = epoch;```
or after 
```
staker.epochLastUnstakedOrFirstStaked = epoch;
```
does it matter ?

adi44: parameters.withdrawlockPeriod() is external call

abhishekvispute: oh okay

abhishekvispute (approved):

hrishikeshio (approved):

PR #361, opened 2021-08-24
staker inactivity penatly bug fix

### LINKED ISSUE #284: staker can avoid inactivity penatly

SamAg19:
Fixes #284

### DIFF
--- a/contracts/Core/StakeManager.sol
+++ b/contracts/Core/StakeManager.sol
@@ -140,6 +140,10 @@ contract StakeManager is Initializable, ACL, StakeStorage, StateManager, Pause {
         require(sAmount > 0, "Non-Positive Amount");
         IStakedToken sToken = IStakedToken(staker.tokenAddress);
         require(sToken.balanceOf(msg.sender) >= sAmount, "Invalid Amount");
+        if (stakerIds[msg.sender] == stakerId) {
+            // Staker can remain inactive for long time and then withdraw the funds before committing to avoid inactivity penalty.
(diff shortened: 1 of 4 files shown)


contracts/Core/StakeManager.sol
```diff
@@ -140,6 +140,10 @@ contract StakeManager is Initializable, ACL, StakeStorage, StateManager, Pause {
         require(sAmount > 0, "Non-Positive Amount");
         IStakedToken sToken = IStakedToken(staker.tokenAddress);
         require(sToken.balanceOf(msg.sender) >= sAmount, "Invalid Amount");
+        if (stakerIds[msg.sender] == stakerId) {
+            // Staker can remain inactive for long time and then withdraw the funds before committing to avoid inactivity penalty.
+            rewardManager.giveInactivityPenalties(epoch, stakerId);
```
abhishekvispute, line 145: Hey 
How its assured penalty wont be applied twice 
lets say I unstake in this epoch, we apply inactivity penalties
I commit in next epoch, then wont this be applied again
https://github.com/razor-network/contracts/blob/eb2bd74a40189f22d60c602654652f874ada522e/contracts/Core/VoteManager.sol#L50

GauravJain9: If staker calls unstake and after that didn't withdraw and after withdraw_lock_period calls resetlock and then commits then penalty will be applied twice ? @abhishekvispute @SamAg19

SkandaBhat:
Closing this as we have a better implementation.

PR #373, opened 2021-08-26
Test for modifier Initializer 

GauravJain9:
#350
Here added thet test case for checking modifier Initializer.

Taken reference from OpenZeppelin. They have written a mock file which has function that uses modifier 'Initializer' & one more function initializeNested which also uses modifier 'initializer'.

https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/proxy/utils/Initializable.sol

https://github.com/OpenZeppelin/openzeppelin-contracts/blob/b9125001f0a1c44d596ca3a47536f1a467e3a29d/contracts/mocks/InitializableMock.sol

https://github.com/OpenZeppelin/openzeppelin-contracts/blob/b9125001f0a1c44d596ca3a47536f1a467e3a29d/test/utils/Initializable.test.js

### DIFF
--- a/contracts/Core/StateManager.sol
+++ b/contracts/Core/StateManager.sol
@@ -4,11 +4,6 @@ pragma solidity ^0.8.0;
 import "./storage/Constants.sol";
 
 contract StateManager is Constants {
-    modifier checkEpoch(uint32 epoch, uint32 epochLength) {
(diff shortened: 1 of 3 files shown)


SkandaBhat (approved):
LGTM.

SamAg19 (approved):

PR #374, opened 2021-08-27
Inactivity bug

### LINKED ISSUE #284: staker can avoid inactivity penatly

### LINKED ISSUE #359: Reset Lock Penalty is calculated on whole stake
**Describe the bug**
As of now, if the delegator or staker does resetLock, the penalty is calculated from the whole stake and then the tokens are burned from the given delegator or staker.

For example.
Total Stake : 50,000 RZR
Delegator sRZR Bal : 5,000 sRZR

Penalty : 50,000 * 1 % = 5,000
then sRZR worth of 5,000 RZR are burned from delegator, lets say its 1 rzr : 1 sRZR

then now 
Delegators balance would be 0 

which is harsh, and not fair, as it's not a 1% penalty its 100 % in this case

this calculation makes penalty proportional to stake, which should not be

**Solution** 
Calculate penalty from lock.amount

SamAg19:
Fixes #284 
Fixes #359 
> problem
> 
>     * can bypass inactivity penalty
>       -unstake() and withdraw() after inactivity since we are not checking inactivity there
> 
>     * after partial unstaking staker still cant participate
>       
>       * after returning after lock period he will get inactivity penalty
> 
> 
> solution 1
> 
>     * unstake() convert unstaked srzr to rzr and reduce from stake; add to lock 10000
> 
>     * withdraw() allow locked rzr to withdraw

> thisStaker.epochFirstStaked -> rename to thisStaker.epochLastUnstakedOrFirstStaked

### DIFF
--- a/contracts/Core/StakeManager.sol
+++ b/contracts/Core/StakeManager.sol
@@ -179,8 +179,6 @@ contract StakeManager is Initializable, ACL, StakeStorage, StateManager, Pause {
         require(lock.withdrawAfter != 0, "Did not unstake");
         require(lock.withdrawAfter <= epoch, "Withdraw epoch not reached");
         require(lock.withdrawAfter + parameters.withdrawReleasePeriod() >= epoch, "Release Period Passed"); // Can Use ResetLock
-        //require(staker.stake > 0, "Nonpositive Stake");
-        //require((voteManager.getCommitmentEpoch(stakerId)) != epoch, "Already commited");
 
         uint256 lockedAmount = lock.amount;
 
--- a/contracts/Core/StakeManager.sol
+++ b/contracts/Core/StakeManager.sol
@@ -144,15 +144,16 @@ contract StakeManager is Initializable, ACL, StakeStorage, StateManager, Pause {
         require(sToken.balanceOf(msg.sender) >= sAmount, "Invalid Amount");
 
         uint256 rAmount = _convertSRZRToRZR(sAmount, staker.stake, sToken.totalSupply());
-        require(sToken.burn(msg.sender, sAmount), "Token burn Failed");
         staker.stake = staker.stake - rAmount;
 
         locks[msg.sender][staker.tokenAddress] = Structs.Lock(rAmount, epoch + (parameters.withdrawLockPeriod()));
 
         staker.epochLastUnstakedOrFirstStaked = epoch;
 
-        emit Unstaked(epoch, stakerId, rAmount, staker.stake, block.timestamp);
         //emit event here
+        emit Unstaked(epoch, stakerId, rAmount, staker.stake, block.timestamp);
+
+        require(sToken.burn(msg.sender, sAmount), "Token burn Failed");
     }
 
     /// @notice staker/delegator can withdraw their funds after calling unstake and withdrawAfter period.
@@ -189,8 +190,8 @@ contract StakeManager is Initializable, ACL, StakeStorage, StateManager, Pause {
         // Check commission rate >0
         if (stakerIds[msg.sender] != stakerId && staker.commission > 0) {
             uint256 commission = (lockedAmount * staker.commission) / 100;
-            require(razor.transfer(staker._address, commission), "couldnt transfer");
             lockedAmount = lockedAmount - commission;
+            require(razor.transfer(staker._address, commission), "couldnt transfer");
         }
 
         //Transfer stake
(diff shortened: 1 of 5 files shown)


SkandaBhat:
@SamAg19 please resolve conflicts.

contracts/Core/StakeManager.sol
```diff
@@ -169,38 +179,26 @@ contract StakeManager is Initializable, ACL, StakeStorage, StateManager, Pause {
...
-        require((voteManager.getCommitmentEpoch(stakerId)) != epoch, "Already commited");
-        if (stakerIds[msg.sender] == stakerId) {
-            // Staker Must not particiapte in withdraw lock period, To counter Hit and Run Attacks
-            require(
-                (lock.withdrawAfter - parameters.withdrawLockPeriod()) >= voteManager.getEpochLastCommitted(stakerId),
-                "Participated in Lock Period"
-            );
-        }
+        //require(staker.stake > 0, "Nonpositive Stake");
+        //require((voteManager.getCommitmentEpoch(stakerId)) != epoch, "Already commited");
```
abhishekvispute, line 184: I think we can remove this two requires now

test/StakeManager.js
```diff
@@ -392,71 +391,6 @@ describe('StakeManager', function () {
       await assertRevert(tx, 'Nonpositive stake');
     });
 
-    it('Staker should not be able to withdraw after withdraw lock period if voted in withdraw lock period', async function () {
```
abhishekvispute, line 395: Hey here can modify this test case, that staker should be able to withdraw even if they have participated in the withdraw lock period.
This will test partial withdraw as well

abhishekvispute: All else looks good

contracts/Core/StakeManager.sol
SamAg19: Removed

test/StakeManager.js
SamAg19: Test case has been added

abhishekvispute (dismissed):
LGTM

contracts/Core/StakeManager.sol
```diff
@@ -245,18 +241,17 @@ contract StakeManager is Initializable, ACL, StakeStorage, StateManager, Pause {
         require(locks[msg.sender][stakers[stakerId].tokenAddress].amount != 0, "Existing Lock doesnt exist");
 
         Structs.Staker storage staker = stakers[stakerId];
-        IStakedToken sToken = IStakedToken(stakers[stakerId].tokenAddress);
-
-        uint256 penalty = (staker.stake * parameters.resetLockPenalty()) / 100;
```
hrishikeshio, line 250: why removing penalty?

contracts/Core/StakeManager.sol
```diff
@@ -245,18 +241,17 @@ contract StakeManager is Initializable, ACL, StakeStorage, StateManager, Pause {
...
+        uint256 lockedAmount = locks[msg.sender][stakers[stakerId].tokenAddress].amount;
+        IStakedToken sToken = IStakedToken(staker.tokenAddress);
 
-        // Converting Penalty into sAmount
-        uint256 sAmount = _convertRZRtoSRZR(penalty, staker.stake, sToken.totalSupply());
+        uint256 penalty = (lockedAmount * parameters.resetLockPenalty()) / 100;
+        lockedAmount = lockedAmount - penalty;
```
hrishikeshio, line 248: what if penalty > lockedAmount?

contracts/Core/StakeManager.sol
```diff
@@ -169,38 +179,24 @@ contract StakeManager is Initializable, ACL, StakeStorage, StateManager, Pause {
...
         _resetLock(stakerId);
 
         // Transfer commission in case of delegators
         // Check commission rate >0
         if (stakerIds[msg.sender] != stakerId && staker.commission > 0) {
-            uint256 commission = (rAmount * staker.commission) / 100;
+            uint256 commission = (lockedAmount * staker.commission) / 100;
             require(razor.transfer(staker._address, commission), "couldnt transfer");
-            rAmount = rAmount - commission;
+            lockedAmount = lockedAmount - commission;
```
hrishikeshio, line 193: what if commission > lockedAmount?

contracts/Core/StakeManager.sol
```diff
@@ -137,11 +137,21 @@ contract StakeManager is Initializable, ACL, StakeStorage, StateManager, Pause {
...
+        rewardManager.giveInactivityPenalties(epoch, stakerId);
+
         IStakedToken sToken = IStakedToken(staker.tokenAddress);
         require(sToken.balanceOf(msg.sender) >= sAmount, "Invalid Amount");
-        locks[msg.sender][staker.tokenAddress] = Structs.Lock(sAmount, epoch + (parameters.withdrawLockPeriod()));
 
-        emit Unstaked(epoch, stakerId, sAmount, staker.stake, block.timestamp);
+        uint256 rAmount = _convertSRZRToRZR(sAmount, staker.stake, sToken.totalSupply());
+        require(sToken.burn(msg.sender, sAmount), "Token burn Failed");
+        staker.stake = staker.stake - rAmount;
```
hrishikeshio, line 148: state changes should be done before interactions

contracts/Core/StakeManager.sol
SamAg19: for penalty > lockedAmount, resetLockPenalty will have to be greater 100 and since resetLockPenalty is in terms of percentage, penalty wont be greater than lockedAmount

contracts/Core/StakeManager.sol
SamAg19: same with commission also

contracts/Core/StakeManager.sol
SamAg19: Done

hrishikeshio (approved):

abhishekvispute (approved):
LGTM

PR #376, opened 2021-08-27
Staker should be allowed to set comission before accepting delegation.

### LINKED ISSUE #372: setcommission should be done before accepting delegation.
setcommission should be done before accepting delegation.
staker can accept delegation and after people delegate, they can set commission too high.

adi44:
fixes #372

### DIFF
--- a/contracts/Core/StakeManager.sol
+++ b/contracts/Core/StakeManager.sol
@@ -224,7 +224,7 @@ contract StakeManager is Initializable, ACL, StakeStorage, StateManager, Pause {
     function setCommission(uint8 commission) external {
         uint32 stakerId = stakerIds[msg.sender];
         require(stakerId != 0, "staker id = 0");
-        require(stakers[stakerId].acceptDelegation, "Delegetion not accpected");
+        require(!stakers[stakerId].acceptDelegation, "Delegation accepted");
         require(stakers[stakerId].commission == 0, "Commission already intilised");
         stakers[stakerId].commission = commission;
     }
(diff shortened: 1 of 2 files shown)


contracts/Core/StakeManager.sol
```diff
@@ -224,7 +225,7 @@ contract StakeManager is Initializable, ACL, StakeStorage, StateManager, Pause {
     function setCommission(uint8 commission) external {
         uint32 stakerId = stakerIds[msg.sender];
         require(stakerId != 0, "staker id = 0");
-        require(stakers[stakerId].acceptDelegation, "Delegetion not accpected");
+        require(!stakers[stakerId].acceptDelegation, "Delegation accepted");
```
abhishekvispute, line 228: Hey @adi44 
Is this require needed anymore ?
Without this well seq is assured
As you cant accept delegation unless you have setCommission
So if you are calling setCommission, its given you have not accepted delegation.

adi44: Yes true, we can remove that. it seems to be redundant.

test/StakeManager.js
```diff
@@ -584,36 +584,21 @@ describe('StakeManager', function () {
...
 
-    it('Staker should not be able to set commission if delegation not accepted', async function () {
-      const tx = stakeManager.connect(signers[1]).setCommission(5);
-      await assertRevert(tx, 'Delegetion not accpected');
+    it('Staker should not be able to accept delegation if comission is not set', async function () {
+      const tx = stakeManager.connect(signers[1]).setDelegationAcceptance('true');
+      await assertRevert(tx, 'comission not set');
     });
-
     it('Delegator should not be able to delegate if delegation not accepted', async function () {
```
abhishekvispute, line 591: Hey 
did we miss adding this test case ?

contracts/Core/StakeManager.sol
abhishekvispute: cool

test/StakeManager.js
adi44: no all the testcases are there. but I had changed few things as per the new require condition

abhishekvispute: cool, got it

abhishekvispute (approved):
LGTM

SkandaBhat (approved):

PR #380, opened 2021-08-31
Delegator should not able to delegate, if staker is inactive

### LINKED ISSUE #366: Check for if staker stakes and later withdraws his stake
**Solution**
Add a check on delegate if epochLastRevealed > current epoch - grace

abhishekvispute:
Fixes #366 

Here as of now, only epochLastRevealed is considered to measure the inactivity 
Whereas in calculateInactivityPenalty we are considering epochUnstakedorFirstStaked as well

The reason behind doing this is

In case of inactivity penalties, it makes sense to have ```epochUnstaked```, as once it is applied, we update ```epochUnstaked``` so in the next epoch, those penalties won't be applied again
So if we see ```epochUnstaked``` actually doesn't represent stakers activity, it only added as a solution to the above case

If we add epochUnstaked here as well, the following comment which pointed this issue will also not hold true
https://github.com/razor-network/contracts/pull/353#discussion_r694719820

Because staker would be considered as active even if they unstacked.

Please let me know if you guys think otherwise, and if we should add it.


contracts/Core/StakeManager.sol
```diff
@@ -345,6 +347,14 @@ contract StakeManager is Initializable, ACL, StakeStorage, StateManager, Pause {
         return stakers[stakerId].epochFirstStaked;
     }
 
+    /// @return isStakerActive : Activity < Grace
+    function isStakerActive(uint32 stakerId, uint32 epoch) public view returns (bool) {
```
abhishekvispute, line 351: This is made public so web client or CLI can access anytime if staker is active.
We can achieve so via indexing in web client as well,
But in CLI via this method, it can first check if given staker is active

Let me know if you guys think its not needed, we can make it internal

SkandaBhat: This makes sense.

SkandaBhat (approved):
Looks good to me.

hrishikeshio (approved):

contracts/Core/StakeManager.sol
```diff
@@ -345,6 +347,14 @@ contract StakeManager is Initializable, ACL, StakeStorage, StateManager, Pause {
...
+        uint32 epochLastRevealed = voteManager.getEpochLastRevealed(stakerId);
+        uint32 inactiveEpochs = (epoch - epochLastRevealed == 0) ? 0 : epoch - epochLastRevealed - 1;
+        if (inactiveEpochs < parameters.gracePeriod()) return true;
+        return false;
```
hrishikeshio, line 355: the three lines above can be combined for more efficient code

return ( (epoch - epochLastRevealed) >= parameters.gracePeriod());

hrishikeshio (changes requested):
.

contracts/Core/StakeManager.sol
abhishekvispute: done https://github.com/razor-network/contracts/pull/380/commits/28f2a75119b2175b53a2f9dad03ac5fea363c509

hrishikeshio (approved):

hrishikeshio (dismissed):

hrishikeshio (approved):

SamAg19 (approved):

PR #388, opened 2021-09-01
removed redundant code in Vote Manager

### LINKED ISSUE #387: Redundant code in Vote Manager
https://github.com/razor-network/contracts/blob/3deec9767e9e189a44ee869c0286d411f2f5c56b/contracts/Core/VoteManager.sol#L112

For the above line we have two functions which are returning the same thing , and this particular function is not being used anywhere.

shekhar2807:
fixes #387 

Removed the redundant code in Vote Manager


test/StakeManager.js
```diff
@@ -566,6 +566,12 @@ describe('StakeManager', function () {
       const tx = stakeManager.connect(signers[1]).setDelegationAcceptance('true');
       await assertRevert(tx, 'comission not set');
     });
+
+    it('Staker should not be able to decreaseCommission if comission is zero', async function () {
```
abhishekvispute, line 570: Hey, this revert doesn't indicate that if the old commission is zero, staker should not be able to decrease it,
It indicates once the commission is set, it cant be decreased to 0 
right?

If so, please update this title

shekhar2807:
Closing this will create two seperate PR's for test cases and redundant code.

PR #407, opened 2021-09-07
Allow clients to request for random no, generate it from stakers secrets and allow client to pull it once ready

### LINKED ISSUE #296: Access random number from randao
TBD

abhishekvispute:
Fixes #296 

A new separate module is created for this called 'randomNumber'
Appropriate comments are added on function only
Please check

_The reason, why it's not added as sep individual job as we discussed last time is 
it would have required changes in dispute and penalties mech as well, 
With this impl its less coupled_

If you guys think we should have it in block only, let me know


hrishikeshio (approved):

abhishekvispute:
Hi Guys 
So as we yesterday discussed on channel, to avoid DOS,  have migrated to bytes32, and client only passes it.
In register, we check if this requestId is unique or not, if its not we revert.

This had created another vulnerability.
Mostly the clients using our randomNo service will have requestId generated on intervals, programmatically only.
So what attacker could do is, request for those requestID in advance,
so DOS would have happened on client side

Hence what we have done is, we have mapped requestID on basis of msg.sender as well
```
requests[msg.sender][requestId] 
```
Avoiding this attack vector


_Now we may say that as we have msg.sender we can go back to uint256 requestId and increment it, yes we can do that, but with this bytes32 client feeded requestID we give client control over salt_

CC : @hrishikeshio @SkandaBhat @SamAg19

contracts/randomNumber/RandomNoManager.sol
```diff
@@ -0,0 +1,93 @@
...
+    /// @param requestId : A unique id per request
+    /// @return random number
+    function getRandomNumber(bytes32 requestId) external view override returns (uint256) {
+        uint32 epochOfRequest = requests[msg.sender][requestId];
+        return _generateRandomNumber(epochOfRequest, requestId);
+    }
+
+    /// @notice Allows client to get generic random number of last epoch
+    /// @return random number
+    function getGenericRandomNumberOfLastEpoch() external view override returns (uint256) {
```
SkandaBhat, line 73: Probably can be called getGenericRandomNumber and the last epoch part can be mentioned in documentation.

SkandaBhat: Just saw `getGenericRandomNumber` Why two methods?

abhishekvispute: one allows to get genericRandomNumber of any epoch, if client wants to , there epoch needs to be passed
one allows to get genericRandomNumber of last epoch, their client doesn't have to pass anything

SkandaBhat: I meant, `getGenericRandomNumberOfLastEpoch` and `getGenericRandomNumber` are pretty much the same, can we have just one `getGenericRandomNumber`? Are these different use cases for getting a random number for this epoch and the previous epoch?

abhishekvispute: hmm yes 

```getGenericRandomNumber```
whenever client starts a game, they can store what current epoch was in their contract 
and at time of pulling lever can call ```getGenericRandomNumber(epoch)```

```getGenericRandomNumberLastEpoch```
This allows use case where random number is irrespective of epoch, just fetch latest random number 
We can say that client can fetch the current epoch on itself and call ```getGenericRandomNumber(epoch-1)```
That way also execution cost remains same, there is no extra execution cost on our side, bothways client needs to do work.
Via separate method, its just better UX

Comment (SkandaBhat) [reply contracts/randomNumber/RandomNoManager.sol:73]: Ok, sounds good.

**Comment by SkandaBhat** [comment]
@abhishekvispute please add the new modifications planned in a comment here.

**Comment by abhishekvispute** [comment]
> @abhishekvispute please add the new modifications planned in a comment here.

Done, Please check new implementation

In an earlier case client had to pass bytes32 requestId which had created two new issues 
**First**, it was not assured client is going to generate it the right way 
as if its duplicate, needlessly it's reverted 
so we are generating requestId for a client now, either way at end it remains same, either client does it on their side or we do, keccak cost need to be spent.

now on generating requestId we are generating it from users nonce and address so it's unique always and is DOS resistant.
plus in getRandomNymber we don't have to map it on the basis of msg.sender anymore, so the dependency of only register calling it is not there anymore, this was the **second*** issue in earlier impl

**Review by hrishikeshio — APPROVED** [review]

File: contracts/randomNumber/RandomNoManager.sol [hunk 2]
```diff
@@ -0,0 +1,88 @@
...
+        parameters = IParameters(parametersAddress);
+    }
+
+    /// @notice Allows Client to register for random number
+    /// Per request a rquest id is generated, which is binded to one epoch
+    /// this epoch is current epoch if Protocol is in commit state, and epoch + 1 if in any other states
+    /// @return requestId : unique request id
+    function register() external override initialized returns (bytes32 requestId) {
+        uint32 epoch = getEpoch(parameters.epochLength());
+        State state = getState(parameters.epochLength());
```
hrishikeshio, line 35: two repeated external calls. Store it in memory instead

contracts/randomNumber/RandomNoManager.sol
```diff
@@ -0,0 +1,88 @@
...
+        nonce[msg.sender]++;
```
hrishikeshio, line 36: avoid unary operator

hrishikeshio (changes requested):
.

contracts/randomNumber/RandomNoManager.sol
abhishekvispute: done

contracts/randomNumber/RandomNoManager.sol
abhishekvispute: done

hrishikeshio (approved):

SkandaBhat (approved):

PR #426, opened 2021-09-13
Delegator

### LINKED ISSUE #340: Delegator is not updated with recent developments.
**Describe the bug**
Delegator is out of date. We need to add getCollection, getResult etc in delegator.

SamAg19:
fixes #340


contracts/Delegator.sol
```diff
@@ -2,19 +2,39 @@
...
+    IParameters public parameters;
     IAssetManager public assetManager;
+    IBlockManager public blockManager;
 
-    function upgradeDelegate(address newDelegateAddress) external onlyRole(DEFAULT_ADMIN_ROLE) {
+    function initialize(
+        address newDelegateAddress,
+        address newResultAddres,
+        address parametersAddress
+    ) external onlyRole(DEFAULT_ADMIN_ROLE) {
```
hrishikeshio, line 22: initializer modifier missing!

hrishikeshio: import "../Initializable.sol"; missing

SamAg19:
Hey guys,
Due to the recents PRs that have been merged, The contract size for Asset Manager has exceeded. I have tried lowering the number of optimization runs but it does not seem to fix the problem. We may need to divide the contract

contracts/Delegator.sol
```diff
@@ -27,10 +45,35 @@ contract Delegator is ACL {
...
+
+    function getNumActiveAssets() external view returns (uint256) {
+        return (assetManager.getActiveAssets()).length;
+    }
+
+    function getActiveAssets() external view returns (uint8[] memory) {
+        return assetManager.getActiveAssets();
+    }
+
+    function getResult(bytes32 _name) public view returns (uint32) {
```
abhishekvispute, line 73: Hey is this possible scenario ?

Epoch x
ActiveAssets [1,2,3]
BlockMedian [10,20,30]

Epoch X + 1
State Confirm : Deactivate 2, ActiveAssets[1,3]
Client calls get result for asset id 3, index we get is 1
when we fetch median from lastBlock we fetch it for 1, where as in last block it was at 2
giving incorrect value

contracts/Delegator.sol
```diff
@@ -27,10 +45,35 @@ contract Delegator is ACL {
...
+        uint8 index = assetManager.getAssetIndex(ids[_name]);
+        uint32 epoch = parameters.getEpoch();
+        uint32[] memory medians = blockManager.getBlockMedians(epoch - 1);
```
abhishekvispute, line 76: As its external call, maybe we can optimise this to only fetch medians for given index, not for all and then pick one

contracts/Delegator.sol
SamAg19: Now thinking about this, the following scenario can be possible

PR #427, opened 2021-09-13
Added enum for stakeChange Reason

adi44:


contracts/Core/RewardManager.sol
```diff
@@ -51,7 +51,7 @@ contract RewardManager is Initializable, ACL, Constants, IRewardManager {
     function giveBlockReward(uint32 stakerId, uint32 epoch) external override onlyRole(REWARD_MODIFIER_ROLE) {
         uint256 blockReward = parameters.blockReward();
         uint256 newStake = stakeManager.getStake(stakerId) + (blockReward);
-        stakeManager.setStakerStake(epoch, stakerId, newStake);
+        stakeManager.setStakerStake(epoch, stakerId, uint8(StakeChanged.BlockReward), newStake);
```
hrishikeshio, line 54: is typecasting required>?

adi44: no we can do without that also

hrishikeshio: remove if not required

hrishikeshio (approved):

SkandaBhat (approved):

PR #429, opened 2021-09-13
Block Insertion Bug

### LINKED ISSUE #414: Possible Redundant If condition in insertApporipriately
In insertAppropriately 
We have one condition 
https://github.com/razor-network/contracts/blob/ef7dd744ece41079257d4591ec7cfc46a24b41ca/contracts/Core/BlockManager.sol#L238

but I am not able to think of any reason, of it becoming true 
because 
```proposedBlocks[epoch][i].biggestInfluence ``` we set in propose
```
   proposedBlocks[epoch][numProposedBlocks] = Structs.Block(proposerId, medians, iteration, biggestInfluence);
   _insertAppropriately(epoch, numProposedBlocks, iteration, biggestInfluence);
```
And same is passed in _insertAppropriately, so I think it always will be equal only

### LINKED ISSUE #428: Proposed Block not being inserted appropriately
**Describe the bug**
Proposed blocks are not being sorted properly based on their iteration due to the issue in the following if condition:
https://github.com/razor-network/contracts/blob/9521cdf965dbe59c2ba414c21417b7041f59bb31/contracts/Core/BlockManager.sol#L242

Here, it is trying to sort serially whereas it should be sorting based on the sortedProposedBlockIds array

### PR DESCRIPTION
**Pull request by SamAg19**
fixes #428 
fixes #414

### CODE REVIEW

File: contracts/Core/BlockManager.sol [hunk 1]
```diff
@@ -239,7 +239,7 @@ contract BlockManager is Initializable, ACL, BlockStorage, StateManager {
                 pushAt = i;
                 break;
             }
-            if (proposedBlocks[epoch][i].iteration > iteration) {
+            if (proposedBlocks[epoch][sortedProposedBlockIds[epoch][i]].iteration > iteration) {
```
hrishikeshio, line 242: line 238 also needs fix

GauravJain9: line 238 actually the whole if block is not required

abhishekvispute: addressed here, its not redundant 
https://github.com/razor-network/contracts/issues/414#issuecomment-918263520

abhishekvispute (dismissed):

test/BlockManager.js
```diff
@@ -896,5 +896,163 @@ describe('BlockManager', function () {
...
+      await razor.transfer(signers[14].address, tokenAmount('421000'));
+      await razor.transfer(signers[15].address, tokenAmount('420000'));
+      await razor.transfer(signers[16].address, tokenAmount('419000'));
+      await razor.transfer(signers[17].address, tokenAmount('418000'));
+
+      // await razor.connect(signers[12]).approve(stakeManager.address, tokenAmount('423000'));
+      // await stakeManager.connect(signers[12]).stake(epoch, tokenAmount('423000'));
+      //
+      // await razor.connect(signers[13]).approve(stakeManager.address, tokenAmount('422000'));
+      // await stakeManager.connect(signers[13]).stake(epoch, tokenAmount('422000'));
```
abhishekvispute, line 914: hey we can remove this comments, if not needed

abhishekvispute (dismissed):

test/BlockManager.js
```diff
@@ -899,5 +899,155 @@ describe('BlockManager', function () {
 
       await assertRevert(tx, 'Proposed Alternate block is identical to proposed block');
     });
+    it('Blocks should be proposed according to iteration', async function () {
```
hrishikeshio, line 902: also write tests about https://github.com/razor-network/contracts/pull/429/files#diff-ef39b70abd99dba15254b650eeab6fe94780b7804fbd0715169e10e8ad67e198R243

SamAg19: Wrote a test case on this

abhishekvispute (approved):

hrishikeshio (approved):

PR #436, opened 2021-09-15
Added Weights for Jobs

### LINKED ISSUE #435: Add weights for assets

**Additional context**

abhishekvispute:
Fixes #435 

With this PR we have added weights to job 

so now whenever the client proposes a median for collection  
they have to do weighted median
```
j1v*j1w + j2v*j2w + j2v*j2w ....
----------------------------------------
	j1w + j2w + j3w...

j(i)v = value of ith job
j(i)w = weight of ith job
````

on setting the weight of the job, have kept condition of max 100,
it's true that value is added in the numerator as well as deno, so one can say there wont be an issue.
it's for setting up limit
lets say
I have assets with weights 30,40,90 as weights 
when you add a new asset in this collection you add it with considering 100 as max, so there is a defined limit
you cant add it with 900 or something, which would dilute other weights

### DIFF
--- a/contracts/Core/AssetManager.sol
+++ b/contracts/Core/AssetManager.sol
@@ -14,6 +14,7 @@ contract AssetManager is ACL, AssetStorage, Constants, StateManager, IAssetManag
     event JobCreated(
         uint8 id,
         JobSelectorType selectorType,
+        uint8 weight,
         int8 power,
         address creator,
         uint32 epoch,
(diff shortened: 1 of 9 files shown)

### CODE REVIEW

File: contracts/Core/AssetManager.sol [hunk 1]
```diff
@@ -11,20 +11,19 @@ import "./ACL.sol";
 contract AssetManager is ACL, AssetStorage, Constants, StateManager, IAssetManager {
     IParameters public parameters;
 
-    event JobCreated(
+    event JobCreated(Structs.Job job, uint256 timestamp);
```
abhishekvispute, line 14: had to modify this to struct from direct elements, as there was stack too deep error
https://app.circleci.com/pipelines/github/razor-network/contracts/562/workflows/8209fb1a-2f02-42ef-ad65-84d95dcf75b5/jobs/3367

abhishekvispute:
> AssetManager	removeJobFromCollection	+0.07	55017.5	54979.5	+38.00

this is the proof of tight packing :)
the newly added param, don't increase any cost, as it occupies same slot

hrishikeshio (approved):

SamAg19 (approved):

PR #443, opened 2021-09-16
Penalty bug

### LINKED ISSUE #392: Staker can avoid inactivity penalties by unstaking a small amount every epoch
**Describe the bug**
Right now, inactivity penalties are given out during unstake and the staker's unstake action is updated which prevents him from getting a penalty when the staker commits in the next epoch.

The staker can bypass inactivity penalties during commit state by unstaking a very small amount of razors every epoch, thereby not taking voting in the network.

### LINKED ISSUE #396: Staker is levied inactivity penalties if she resets lock
staker stakes 11k
Unstakes 10.5k, resulting stake going 500
she cannot commit, reveal during lock period
she misses the withdraw release period, not being able to withdraw
Resets lock, pays the penalty and locked amount is added back to stake.
At the time unstaking, she is levied inactivity penalties

### LINKED ISSUE #445: Staker cant resetLock if the staker wants to unstake his entire stake and then misses withdraw release period
**Describe the bug**
If the staker unstakes his entire stake and misses the withdraw release period, He wont be able to reset lock and reinstate his stake due to the require statement in ` _convertRZRtoSRZR()`

https://github.com/razor-network/contracts/blob/81503fd9a8ecc0f5b5f9a5ebd55b51aa7c5a73c9/contracts/Core/StakeManager.sol#L267

https://github.com/razor-network/contracts/blob/81503fd9a8ecc0f5b5f9a5ebd55b51aa7c5a73c9/contracts/Core/StakeManager.sol#L431

SamAg19:
fixes #392 
fixes #396 
fixes #445


contracts/Core/StakeManager.sol
```diff
@@ -267,6 +266,9 @@ contract StakeManager is Initializable, ACL, StakeStorage, StateManager, Pause,
         uint256 sAmount = _convertRZRtoSRZR(lockedAmount, staker.stake, sToken.totalSupply());
 
         //Updating Staker Stake
+        if (staker.stake < parameters.minStake()) {
```
abhishekvispute, line 269: Hey, why here this min stake condition is added, why it needs to be lesser than minStake ?

contracts/lib/Structs.sol
```diff
@@ -18,7 +18,7 @@ library Structs {
         address tokenAddress;
         uint32 id;
         uint32 age;
-        uint32 epochLastUnstakedOrFirstStaked;
+        uint32 epochStakedOrLastPenalized;
```
abhishekvispute, line 21: Hey, should we add 'first' before 'staked' here, becuase last time it was like this only, and then we added firstStaked as we thought thats proper name

abhishekvispute: this pr : https://github.com/razor-network/contracts/pull/331

contracts/Core/StakeManager.sol
SamAg19: Reason this was done because if the staker's stake is greater than minStake, then it means the staker can still vote in the network and hence, there is no need for his activity to be updated when he resets lock otherwise he may get away with the inactivity penalties when he commits. If the stake is less than minStake, the staker wont be able to vote in the network and therefore, it is important that his activity is updated when he resets lock.

contracts/lib/Structs.sol
SamAg19: Fixed https://github.com/razor-network/contracts/pull/443/commits/994f7726590db47175c2497b2b648bcebd1feae4

contracts/Core/StakeManager.sol
abhishekvispute: okay cool got it 
but I think its very extreme edge case.
without it the outcome of it , is staker who unstaked more then (stake-minStake), and does resetLock getting penalised for both resetLock and inactivity.

abhishekvispute (dismissed):

contracts/Core/StakeManager.sol
abhishekvispute: CC : @hrishikeshio @SkandaBhat

hrishikeshio: This creates a vulnerability where staker goes inactive for long time and unstakes when everyone else has unstaked and his stake goes below minstake. They will avoid inactivity penalty

hrishikeshio: I think we should remove this.
If staker's stake is going below minstake then he can unstake to avoid penalty.
if he is not unstaking that means he is offline and he should be penalized.

SamAg19: removed https://github.com/razor-network/contracts/pull/443/commits/74636206e264a357c578eaff329f19b6d2a137e8

abhishekvispute:
Hey @SamAg19 
I think we are not solving #396 anymore
right ?

SamAg19:
> Hey @SamAg19 I think we are not solving #396 anymore, as that minStake condition is removed right ?

Yep, since we are going to just extend the lock release period, #396 is not an issue anymore

hrishikeshio:
Is this pr ready to merge?

abhishekvispute:
> > Hey @SamAg19 I think we are not solving #396 anymore, as that minStake condition is removed right ?
> 
> Yep, since we are going to just extend the lock release period, #396 is not an issue anymore

Cool please remove that issue from PR desc then 
and lets close it manually

abhishekvispute (approved):

SamAg19:
> > > Hey @SamAg19 I think we are not solving #396 anymore, as that minStake condition is removed right ?
> > 
> > 
> > Yep, since we are going to just extend the lock release period, #396 is not an issue anymore
> 
> Cool please remove that issue from PR desc then and lets close it manually

I think that extension of lock can actually be the solution of #396 since this idea of implementation came up during the discussions of minStake in resetLock which was part of the earlier solution to solve #396

abhishekvispute (dismissed):

contracts/Core/Parameters.sol
```diff
@@ -82,9 +82,9 @@ contract Parameters is ACL, Constants, IParameters {
         withdrawReleasePeriod = _withdrawReleasePeriod;
     }
 
-    function setResetLockPenalty(uint8 _resetLockPenalty) external onlyRole(DEFAULT_ADMIN_ROLE) {
-        emit ParameterChanged(msg.sender, "resetLockPenalty", resetLockPenalty, _resetLockPenalty, block.timestamp);
-        resetLockPenalty = _resetLockPenalty;
+    function setextendLockPenalty(uint8 _extendLockPenalty) external onlyRole(DEFAULT_ADMIN_ROLE) {
```
SkandaBhat, line 85: camelcase is not proper @SamAg19

SkandaBhat (changes requested):

SkandaBhat (approved):
LGTM

contracts/Core/StakeManager.sol
```diff
@@ -246,27 +245,16 @@ contract StakeManager is Initializable, ACL, StakeStorage, StateManager, Pause,
...
-        //Updating Staker Stake
-        staker.stake = staker.stake + lockedAmount;
-
-        _resetLock(stakerId);
-
-        require(sToken.mint(msg.sender, sAmount, lockedAmount));
+        //Giving out the extendLock penalty
+        uint256 penalty = (lock.amount * parameters.extendLockPenalty()) / 100;
+        lock.amount = lock.amount - penalty;
+        lock.withdrawAfter = parameters.getEpoch();
```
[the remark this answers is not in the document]
SamAg19: Then the staker will have to wait for ```currentEpoch + withdrawReleasePeriod``` to withdraw due to this condition https://github.com/razor-network/contracts/blob/82db070d2390b3fc9a2bb05e7bf1889b5b4282e3/contracts/Core/StakeManager.sol#L199
One thing i can add here is that you can only extend lock if you have passed the release period, a condition which is not present currently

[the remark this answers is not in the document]
abhishekvispute: > Then the staker will have to wait for currentEpoch + withdrawReleasePeriod to withdraw due to this condition

yep that's what we mean by extending lock, if we keep current epoch
It won't be extend, its would be like you can withdraw with penalty
not sure if extending is needed, is withdraw with penalty expose us to any issue ?
if not 
we don't even need to have extendLock, why create lock with current epoch.

hmm coming to adding condition, we can do so
earlier we hadn't included it as resetLock could be used in case when delegator lost their sRZR.
now if we are locking in RZR only, its not needed 
so we can add that check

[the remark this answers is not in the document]
abhishekvispute: hmm cool yep it works out 
we don't have to make user wait to withdraw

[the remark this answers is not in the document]
abhishekvispute: User calls withdraw 
It fails as lock expired 
Then user call extendLock
We deduct penalty 
User have currentEpoch + WithdrawReleasePeriod to call Withdraw

cool 👍 

Lets add that check only then, that you mentioned above

abhishekvispute (dismissed):

abhishekvispute (approved):

hrishikeshio (approved):

PR #446, opened 2021-09-16
Inactivity Penalty no more given if stake less than min stake

### LINKED ISSUE #397: Should we render penalty if stake is lesser than min stake?
Should lock penalty and inactivity penalty be applied if stake < minstake?

SamAg19:
fixes #397

### DIFF
--- a/test/StakeManager.js
+++ b/test/StakeManager.js
@@ -1416,5 +1416,35 @@ describe('StakeManager', function () {
       staker = await stakeManager.getStaker(4);
       assertBNEqual(prevStake, staker.stake, 'Inactivity penalties have been levied');
     });
+    it('should not levy inactivity penalities if stake less than minStake ', async function () {
+      await mineToNextEpoch();
+      let epoch = await getEpoch();
+      const stake = tokenAmount('10000');
+      await razor.transfer(signers[16].address, stake);
+      await razor.connect(signers[16]).approve(stakeManager.address, stake);
+      await stakeManager.connect(signers[16]).stake(epoch, stake);
+      const stakerId = await stakeManager.stakerIds(signers[16].address);
+      await mineToNextEpoch();
+
+      epoch = await getEpoch();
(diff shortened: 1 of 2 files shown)


test/StakeManager.js
```diff
@@ -1416,5 +1416,35 @@ describe('StakeManager', function () {
...
+      await mineToNextEpoch();
+
+      epoch = await getEpoch();
+      await stakeManager.setStakerStake(epoch, stakerId, 1, tokenAmount('500'));
+
+      const epochsJumped = GRACE_PERIOD + 4;
+      for (let i = 0; i < epochsJumped; i++) {
+        await mineToNextEpoch();
+      }
+
```
abhishekvispute, line 1436: hey, once grace period is over, inactivity penalties are applied only when you commit or unstake
this test would have hold true without this pr as well.
right ?
I think we should do unstable here

SamAg19: I have unstaked at L1445

abhishekvispute: cool

abhishekvispute (approved):

hrishikeshio:
I don't think it fixes the issue.

It only checks the current minstake.

T1
Stake 20000
Minstake 10000

Staker goes down for 1000 epochs and stays down

T2
Stake 20000
Minstake 30000

Staker withdraws without inactivity penalty.

SamAg19:
> I don't think it fixes the issue.
> 
> It only checks the current minstake.
> 
> T1
> Stake 20000
> Minstake 10000
> 
> Staker goes down for 1000 epochs and stays down
> 
> T2
> Stake 20000
> Minstake 30000
> 
> Staker withdraws without inactivity penalty.

How about we give out inactivity penalties to all the stakers just before we change minStake when we call setMinStake function in Parameters? The gas costs will be incurred by us here as the admin can only call this function

hrishikeshio:
Lets discuss in the issue.

PR #450, opened 2021-09-16
Customised Bounty and Amount to be burned

### LINKED ISSUE #389: Customize amount of bounty, amount to burn when slashed
Currently half of slashed stake is rewarded to bounty hunter.

this incentivises staker to slash self since their stake can be minimal compared to the delegated stake.

abhishekvispute:
Fixes #389 

Have made penalty divider customizable
Earlier it was 2

so slashPenalty/2 was awarded
now its slashPenalty * bountyNum / bountyDenom

have kept bountyNum = 500
bontyDenom = 10000

so at last its slashPenalty/20 , 5 %

from remaining 95% amount, 
```
95% amount * burnSlashNum/burnSlashDenom
```

is burned, remaining stays is contract

### DIFF
--- a/contracts/Core/StakeManager.sol
+++ b/contracts/Core/StakeManager.sol
@@ -298,12 +298,10 @@ contract StakeManager is Initializable, ACL, StakeStorage, StateManager, Pause, 
         address bountyHunter
     ) external override onlyRole(STAKE_MODIFIER_ROLE) returns (uint32) {
         uint256 _stake = stakers[stakerId].stake;
-        (uint16 slashPenaltyNum, uint16 slashPenaltyDenom) = parameters.getSlashNumDenom();
-        uint256 slashPenaltyAmount = (_stake * slashPenaltyNum) / slashPenaltyDenom;
+        uint256 slashPenaltyAmount = (_stake * parameters.slashPenaltyNum()) / parameters.slashPenaltyDenom();
         _stake = _stake - slashPenaltyAmount;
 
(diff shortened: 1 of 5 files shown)


contracts/Core/StakeManager.sol
```diff
@@ -300,21 +300,22 @@ contract StakeManager is Initializable, ACL, StakeStorage, StateManager, Pause,
         uint256 _stake = stakers[stakerId].stake;
         uint256 slashPenaltyAmount = (_stake * parameters.slashPenaltyNum()) / parameters.slashPenaltyDenom();
         _stake = _stake - slashPenaltyAmount;
-        uint256 halfPenalty = slashPenaltyAmount / 2;
 
-        if (halfPenalty == 0) return 0;
+        uint256 bounty = (slashPenaltyAmount * parameters.bountyNum()) / parameters.bountyDenom();
```
abhishekvispute, line 304: **Possible Optimisation**

Guys, 
If we see slashPenaltyNum, denom, bountyNum, denom, burnSlashNum, denom
all are being used here only, so theoretically we can combine all of them into one external call rather than 6.
I tried doing so, please check first commit, issue is with that, it reverts with ```stack to deep``` error if we do it for all 6.

now to resolve this we can combine them all in one structure, or have have only divider, rather than num or denom.
Let me know if you think we should address this.

hrishikeshio (approved):

SamAg19 (approved):

PR #463, opened 2021-09-23
staker should be able to increase stake with n number of RZR token

### LINKED ISSUE #459: Current stake + new stake should be > minstake
https://github.com/razor-network/contracts/blob/81503fd9a8ecc0f5b5f9a5ebd55b51aa7c5a73c9/contracts/Core/StakeManager.sol#L72

adi44:
Currently If staker wants to increase stake, staker needs to stake RZR tokens >= minStake. But now staker can stake any number of tokens untill amount + currentStake >= minStake.

Let Say staker A has currentStake = 990 RZR tokens, amount it want to stake is 10 RZR tokens, then staker can easily increase stake with 10 RZR tokens because it statisfies the condition. 

Fixes #459

### DIFF
--- a/contracts/Core/StakeManager.sol
+++ b/contracts/Core/StakeManager.sol
@@ -69,8 +69,8 @@ contract StakeManager is Initializable, ACL, StakeStorage, StateManager, Pause, 
         checkEpochAndState(State.Commit, epoch, parameters.epochLength())
         whenNotPaused
     {
-        require(amount >= parameters.minStake(), "staked amount is less than minimum stake required");
         uint32 stakerId = stakerIds[msg.sender];
+        require(amount + stakers[stakerId].stake >= parameters.minStake(), "staked amount is less than minimum stake required");
         emit Staked(msg.sender, epoch, stakerId, stakers[stakerId].stake, block.timestamp);
 
         if (stakerId == 0) {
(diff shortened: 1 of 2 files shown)


abhishekvispute (approved):

contracts/Core/StakeManager.sol
```diff
@@ -69,8 +69,8 @@ contract StakeManager is Initializable, ACL, StakeStorage, StateManager, Pause,
         checkEpochAndState(State.Commit, epoch, parameters.epochLength())
         whenNotPaused
     {
-        require(amount >= parameters.minStake(), "staked amount is less than minimum stake required");
         uint32 stakerId = stakerIds[msg.sender];
+        require(amount + stakers[stakerId].stake >= parameters.minStake(), "staked amount is less than minimum stake required");
```
hrishikeshio, line 73: i think it will be more gas efficient and cleaner if the require is put in `if` and `else` blocks instead.

adi44: In that case will it not need to require. One to check if it is new staker then it should stake should be greater than minstake. And other one while increasing stake + amount should be greater than or equal to minstake. Or else we can use one require, if staker is new it should stake atleast minStake

SkandaBhat: Please implement the suggestion.

hrishikeshio (approved):

SamAg19 (approved):

SamAg19 (approved):

abhishekvispute (approved):

PR #464, opened 2021-09-23
Added test cases for newly added parameters in parameters.sol

shekhar2807:
Some new pararmetres were added in parameters.sol so I have tested them to increase the coverage.


hrishikeshio (approved):

SkandaBhat (approved):
Looks good to me.

PR #475, opened 2021-09-28
removed Job Status : Now Job will not have any activity Status.

### LINKED ISSUE #449: Remove activity status from job
- Remove activity status from job.
- If we are removing the last job in the collection, disable collection automatically.

adi44:
In this PR, I have removed the status variable for job. Now only collections will have active Status. 
Renamed function names setAssetStatus --> setCollectionStatus , getActiveStatus --> getCollectionStatus
fixes #449

### DIFF
--- a/contracts/Core/AssetManager.sol
+++ b/contracts/Core/AssetManager.sol
@@ -24,8 +24,6 @@ contract AssetManager is ACL, AssetStorage, Constants, StateManager, IAssetManag
         string url
     );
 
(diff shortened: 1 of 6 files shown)


abhishekvispute (approved):
LGTM 
@SamAg19 as we discussed in yesterdays call, when you will be cleaning this contract.
Please make sure only one naming persist, asset or collection 
as both are same now. 
setAssetStatus or setCollectionStatus , or
getActiveCollections or getActiveAssets 
both will be same

hrishikeshio (approved):

Issue #474, opened 2021-09-27
Slashing the block proposer if they propose incorrect biggestinfluenceID


SkandaBhat:
**Is your feature request related to a problem? Please describe.**

**Describe the solution you'd like**

**Describe alternatives you've considered**

**Additional context**

hrishikeshio:
Is this really an issue?
What is the potential attack? @abhishekvispute

abhishekvispute:
hmm, yes there is attack vector 
we are not reverting for smaller influence, 
so if the block with better itr is proposed with smaller influence, it will be added first on basis of itr.
and if no block is proposed after it to replace it, it stays as chosen block to be confirmed.
that's why we need option to dispute it.

CC : @hrishikeshio

hrishikeshio:
maybe better to track biggest influence in the epoch and compare/update it on every propose. wont consume alot of gas. and less disrputive/risky.

maybe here https://github.com/razor-network/contracts/pull/483/files#diff-ef39b70abd99dba15254b650eeab6fe94780b7804fbd0715169e10e8ad67e198R70

abhishekvispute:
yes, agree, we can do so, shall we do it current pending pr only,
Its a small addition

PR #483, opened 2021-10-03
Replace for larger biggestInflunce in propose

### LINKED ISSUE #467: If biggestInfluencer is incorrect, delete the block
Currently if biggest influencer in proposed block is incorrect, we just move it further in the priority queue 

https://github.com/razor-network/contracts/blob/master/contracts/Core/BlockManager.sol#L247

However, the block is proven to be invalid since it provided incorrect biggestInfluencer. Hence the block should be replaced by the new proposed block.

In a separate issue, we can consider slashing the incorrect block proposer

abhishekvispute:
Fixes #467 

If a block with better biggest influence is proposed it replaces existing one


contracts/Core/BlockManager.sol
```diff
@@ -67,10 +68,10 @@ contract BlockManager is Initializable, ACL, BlockStorage, StateManager, IBlockM
         require(medians.length == assetManager.getNumActiveAssets(), "invalid block proposed");
 
         uint256 biggestInfluence = stakeManager.getInfluence(biggestInfluencerId);
-        uint8 numProposedBlocks = uint8(sortedProposedBlockIds[epoch].length);
+        if (sortedProposedBlockIds[epoch].length == 0) numProposedBlocks = 0;
```
abhishekvispute, line 71: had to move from 
```
numProposedBlocks = uint8(sortedProposedBlockIds[epoch].length);
````
to having numProposedBlocks as state variable as 
if we replace the value,
length of sortedProposedBlockId don't change
so if new block comes in, it would have had same id 

for ex,
lets say we have 1,0
2 is proposed 
it replaces 1
so 2,1
now size is 2
so if we would have had length of array only as numProposedBlocks, new id would have been again 2

I think this problem existed earlier as well, with poping once we reach maxAltBlocks 
we have 
3,4,1,0,2

5 is proposed 

so 2 is popped
its now
3,4,5,1,0

now once new block is proposed here, with previous code,
its id would have been again 5, right ?

File: contracts/Core/BlockManager.sol [hunk 2]
```diff
@@ -256,27 +256,37 @@ contract BlockManager is Initializable, ACL, BlockStorage, StateManager, IBlockM
...
         }
 
-        uint8 pushAt = uint8(sortedProposedBlockIds[epoch].length);
-        for (uint8 i = 0; i < sortedProposedBlockIds[epoch].length; i++) {
-            if (proposedBlocks[epoch][sortedProposedBlockIds[epoch][i]].biggestInfluence < biggestInfluence) {
-                pushAt = i;
+        for (uint8 i = 0; i <= sortedProposedBlockIds[epoch].length; i++) {
+            // Push at last, have worst itr
+            if (i == sortedProposedBlockIds[epoch].length && sortedProposedBlockIds[epoch].length < parameters.maxAltBlocks()) {
+                sortedProposedBlockIds[epoch].push(blockId);
```
hrishikeshio, line 262: if biggest influencer is incorrect, attacker can escape the check here.

contracts/Core/BlockManager.sol
```diff
@@ -256,27 +256,37 @@ contract BlockManager is Initializable, ACL, BlockStorage, StateManager, IBlockM
...
                 break;
             }
-            if (proposedBlocks[epoch][sortedProposedBlockIds[epoch][i]].iteration > iteration) {
-                pushAt = i;
+            // Replace : New Block has better biggest influence
+            else if (proposedBlocks[epoch][sortedProposedBlockIds[epoch][i]].biggestInfluence < biggestInfluence) {
```
hrishikeshio, line 266: sortedProposedBlockIds[epoch][i] will cause index out of bounds when i==sortedProposedBlockIds[epoch].length

contracts/Core/BlockManager.sol
hrishikeshio: Please simplify logic if possible. Not intuitive and hard to debug.

abhishekvispute: hmm sure 
here attacker cant escape 
as this condition is checked at the end of loop

so in between if its found that its incorrect biggestInfluence it reverts

contracts/Core/BlockManager.sol
abhishekvispute: hmm I thought that first if wont allow to reach here 
but it would reach once blocks > maxAltBlocks 
cool will resolve this

test/BlockManager.js
```diff
@@ -1076,7 +1076,7 @@ describe('BlockManager', function () {
       }
     });
 
-    it('Blocks should be sorted properly based on biggestInfluence and iteration', async function () {
+    it('If Biggest Influence of subquecent, block is larger; it should replace, if smaller; should revert', async function () {
```
hrishikeshio, line 1079: Did you remove the test logic regarding iteration based sorting?

test/BlockManager.js
```diff
@@ -1144,99 +1144,43 @@ describe('BlockManager', function () {
...
-        sortedBlockIds.push(i);
-        for (let j = sortedBlockIds.length - 1; j > (pushAt); j--) {
-          sortedBlockIds[j] = sortedBlockIds[j - 1];
-        }
-        sortedBlockIds[pushAt] = i;
-      }
-
-      for (let i = 0; i < proposedBlocksLength; i++) { // 20341
-        sortedProposedBlockId = await blockManager.sortedProposedBlockIds(epoch, i);
-        assertBNEqual(sortedBlockIds[i], sortedProposedBlockId, 'Not sorted properly');
```
hrishikeshio, line 1238: ???

abhishekvispute: we have it checked it one test written just above

contracts/Core/BlockManager.sol
hrishikeshio: true. but gas intensive to compare biggestinfluence with every block.
Not required if proposer can be slashed in dispute by proving higher biggestInfluence.
Which is missing and important.

contracts/Core/BlockManager.sol
hrishikeshio: true. but gas intensive to compare biggestinfluence with every block.
https://github.com/razor-network/contracts/pull/483/files#diff-ef39b70abd99dba15254b650eeab6fe94780b7804fbd0715169e10e8ad67e198R271

Unnecessary if proposer can be slashed in dispute by proving higher biggestInfluence.
Which is missing and important.

abhishekvispute: hmm yep 
okay cool will remove it, even if block with smaller influcence is passed, it will be included as of now.
Then in solution of #474, we will add it.

contracts/Core/BlockManager.sol
abhishekvispute: Fixed https://github.com/razor-network/contracts/pull/483/commits/226903825512b9087390fa4d297389f3502cccdb

contracts/Core/BlockManager.sol
abhishekvispute: Fixed https://github.com/razor-network/contracts/pull/483/commits/226903825512b9087390fa4d297389f3502cccdb

hrishikeshio (approved):

SamAg19 (approved):

PR #486, opened 2021-10-05
Added-amount-for-delegator

adi44:
Now delegated event has extra parameter called amount, that gives the value of razor tokens delegated.


hrishikeshio:
what happened to gas bot

PR #511, opened 2021-10-14
Parameters segregated with governance contract

### LINKED ISSUE #500: Dissolve Parameters and create governance contract that sets parameters in respective contracts.

abhishekvispute:
Fixes #500 
Have segregated parameters into respective childs, and added a governance contract which rules them all.


contracts/Core/parameters/Governance.sol
```diff
@@ -0,0 +1,133 @@
...
+
+    function setEpochLength(uint16 _epochLength) external initialized onlyRole(GOVERNER_ROLE) {
+        emit ParameterChanged(msg.sender, "epochLength", _epochLength, block.timestamp);
+        blockManagerParams.setEpochLength(_epochLength);
+        rewardManagerParams.setEpochLength(_epochLength);
+        stakeManagerParams.setEpochLength(_epochLength);
+        voteManagerParams.setEpochLength(_epochLength);
+        assetManagerParams.setEpochLength(_epochLength);
+        delegatorParams.setEpochLength(_epochLength);
+        randomNoManagerParams.setEpochLength(_epochLength);
```
abhishekvispute, line 131: I agree, that this is not neat, 
as epochLength is being used by StateManager only
so one can say we should have it as param there only.

issue is with dimond inheritance.

if we inherit same GovernanceACL.sol in stateManager then in any main contract, two copies of them get inherited 
so we have to go this way.
one from stateManager and one from their own param child
let me know if there is a way

contracts/Core/parameters/Governance.sol
```diff
@@ -0,0 +1,134 @@
...
+import "./interfaces/IBlockManagerParams.sol";
+import "./interfaces/IRewardManagerParams.sol";
+import "./interfaces/IStakeManagerParams.sol";
+import "./interfaces/IVoteManagerParams.sol";
+import "./interfaces/IAssetManagerParams.sol";
+import "./interfaces/IDelegatorParams.sol";
+import "./interfaces/IRandomNoManagerParams.sol";
+import "../ACL.sol";
+
+// slither-disable-next-line missing-inheritance
```
abhishekvispute, line 13: Disabled as slither is suggesting to have params interfaces to be inherited here
Though function signatures are same, meaning is diff
also two interfaces are going to have some common functions

contracts/Core/parameters/child/AssetManagerParams.sol
```diff
@@ -0,0 +1,13 @@
+// SPDX-License-Identifier: UNLICENSED
+pragma solidity ^0.8.0;
+import "../interfaces/IAssetManagerParams.sol";
+import "./utils/GovernanceACL.sol";
+
+abstract contract AssetManagerParams is GovernanceACL, IAssetManagerParams {
+    uint16 public epochLength = 300;
+
+    function setEpochLength(uint16 _epochLength) external override onlyGovernance {
+        // slither-disable-next-line events-access
```
abhishekvispute, line 10: disabled across all params childs
as they are being called by governance contract only 
and their before setting, we are emitting event

adi44: https://github.com/crytic/slither/wiki/Detector-Documentation#missing-events-arithmetic. actually you need to mention check name which is event-maths

contracts/Core/AssetManager.sol
```diff
@@ -30,8 +29,9 @@ contract AssetManager is ACL, AssetStorage, Constants, StateManager, IAssetManag
 
     event CollectionUpdated(uint8 id, uint32 epoch, uint32 aggregationMethod, int8 power, uint8[] updatedJobIDs, uint256 timestamp);
 
-    constructor(address parametersAddress) {
-        parameters = IParameters(parametersAddress);
+    constructor(address governanceAddress) {
```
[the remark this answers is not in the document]
abhishekvispute: sure will update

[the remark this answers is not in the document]
abhishekvispute: hmm
here in AssetManager initialize is not need theoretically 
so if we add it for pattern 
its going to increase gas cost for all functions of it 
as per function initilized modifier will need to be added 
so should we do it ?

CC : @hrishikeshio

contracts/Core/parameters/Governance.sol
abhishekvispute: done https://github.com/razor-network/contracts/pull/511/commits/a20ff813fa8e2eecd733bec425bcbb06be541e6d

PR #515, opened 2021-10-18
Random Lib made internal

### LINKED ISSUE #509: Make random library functions internal
Make the functions internal.

0xcuriousapple:
Fixes #509

### DIFF
--- a/contracts/lib/Random.sol
+++ b/contracts/lib/Random.sol
@@ -3,12 +3,16 @@ pragma solidity ^0.8.0;
 
 library Random {
     // pseudo random number generator based on hash. returns 0 -> max-1
+    // slither ignore reason : Internal library
+    // slither-disable-next-line dead-code
     function prng(uint256 max, bytes32 randHash) internal pure returns (uint256) {
         uint256 sum = uint256(randHash);
         return (sum % max);
     }
 
     // pseudo random hash generator based on hashes.
+    // slither ignore reason : Internal library
+    // slither-disable-next-line dead-code
     function prngHash(bytes32 seed, bytes32 salt) internal pure returns (bytes32) {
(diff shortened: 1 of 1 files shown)


hrishikeshio (approved):

contracts/lib/Random.sol
```diff
@@ -3,13 +3,17 @@ pragma solidity ^0.8.0;
 
 library Random {
     // pseudo random number generator based on hash. returns 0 -> max-1
-    function prng(uint256 max, bytes32 randHash) external pure returns (uint256) {
+    // slither ignore reason : Internal library
+    // slither-disable-next-line dead-code
+    function prng(uint256 max, bytes32 randHash) internal pure returns (uint256) {
```
hrishikeshio, line 8: i don't understand. how does this fix the issue?

abhishekvispute: hmm slither one or internal functions one ?

Internal Functions 
If we have them as external like before, a library is deployed separately and then its linked to main contract, this saves space for main contract, but their are external calls,  
when we make it internal, its part of bytecode only, hence there are no external calls

On Slither 
Slither says that no other function is using this internal function in this file, so it must be redundant
what's its not understanding maybe is, this is internal library, hence we have to ignore it 
https://github.com/crytic/slither/wiki/Detector-Documentation#dead-code

SkandaBhat: RE: Tenderly, since external lib functions change the deployed bytecode, there was a bytecode mismatch on tenderly verification and push. Since internal functions are embedded, there is no change to the bytecode.

Tenderly should ideally fix this by matching compiled bytecode instead of deployed bytecode, but till then, this change should work.

It also optimises gas fees a little bit so we're achieving multiple things with the PR.

SkandaBhat (approved):

PR #527, opened 2021-10-27
Stakers and delegators freed

### LINKED ISSUE #523: Allow stake, delegate, withdraw in any state
Add more info here.

0xcuriousapple:
Fixes #523 

Stake, Delegate, Withdraw can happen anytime (withdraw can happen only once lock period is over)
Unstake can happen in Commit, Reveal and Confirm 

CC : @yohanelly95 @rajkharvar @ashish10677 

Further on node side
Once this PR is merged use getInfluenceSnapshot everywhere 
weather it is finding iteration or biggestInfluence 
IMO getInfluence is not going to be used on node side anymore 
but let us know if you think its needed anywhere 
If possible please add us as reviewer once you make above PR.

CC : @ashish10677 @Yashk767

### DIFF
--- a/test/StakeManager.js
+++ b/test/StakeManager.js
@@ -1489,12 +1489,12 @@ describe('StakeManager', function () {
       // Propose
       const stakerIdAcc = await stakeManager.stakerIds(signers[4].address);
       const tx = stakeManager.connect(signers[4]).unstake(epoch, stakerIdAcc, 1);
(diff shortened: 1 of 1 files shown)


abhishekvispute:
hmm we are reducing stake in unstake only 
so attacker can propose wrong block 
and unstake 
so later he cant be penalised 
so maybe we will have to restrict unstake not possible in propose and dispute state

PR #540, opened 2021-10-29
Scenario tests

### LINKED ISSUE #455: 100 epochs of constant voting and participation

### LINKED ISSUE #456: Staker is initially active and then becomes inactive more than the GRACE PERIOD

### LINKED ISSUE #532: Staker stakes and participate in network and then unstakes his some amount , which makes his stake less than the min stake and now by using the setMinStake() min stake is changed then staker should be able toparticipate in network again.

### LINKED ISSUE #533: Staker partcipating in network and then remains inactive for long time such that it's stake becomes less than the min stake , then he should not be able to participate.

### LINKED ISSUE #534: Staker participates in network and delegator also delegats also some of it's stake but due to some peanalties delegator withdraws it's stake leading to stake less than min stake of staker , staker should not be able to participate again .

### LINKED ISSUE #535: Staker participates in network and then ustakes some amount and then particpate again and increase it's stake again , everything should work properly.

### LINKED ISSUE #536: Staker unstakes, currently it's in withdraw lock period. there is change in governance parameter and withdraw lock period is now reduced. Will staker be able withdraw early or it remains same.

### LINKED ISSUE #537: Staker unstakes, currently it's in withdraw release period. there is change in governance parameter and withdraw release period is now reduced. Will staker be able withdraw early or it remains same.

### LINKED ISSUE #538: Front-Run Unstake Call

### LINKED ISSUE #539: Front-Run Recurring Unstake Call

### LINKED ISSUE #548: Staker proposes a block and after that blockReward changes (check for changed before and after receiving reward both) so blockReward must be given correctly (according to latest one).

### LINKED ISSUE #549: Staker commits having stake just slightly more than minStake and then minStake becomes more such that it beomes more than the current Stake of the staker and then staker reveals and will not be able to reveal.

### LINKED ISSUE #550: Staker commits having stake just slightly more than reveals and then minStake minStake becomes more such that it beomes more than the nowStake of the staker and then staker proposes and will not be able to propose.

### LINKED ISSUE #551: Staker Gets slashed and its stake becomes below current MinStake, but soon we change minStake and the staker which gets slashed is able to participate again because its stake becomes > minStake.

### LINKED ISSUE #552: After blockReward changes (check for changed before and after receiving reward both), staker must be given the randao penalty correctly (according to latest one).

SamAg19:
fixes #539 
fixes #538 
fixes #537 
fixes #536 
fixes #535 
fixes #534 
fixes #533 
fixes #532 
fixes #456 
fixes #455 
fixes #548 
fixes #549 
fixes #550 
fixes #551 
fixes #552


SkandaBhat (dismissed):
LGTM

test/helpers/utils.js
```diff
@@ -77,6 +77,14 @@ const getEpoch = async () => {
   return blockNumber.div(EPOCH_LENGTH).toNumber();
 };
 
+const getVote = async (medians) => {
```
abhishekvispute, line 80: this is to have random votes ?
this is awesome addition

SamAg19: Yes
Helps in accurate representation of how the network will behave with real votes

test/scenarios.js
```diff
@@ -0,0 +1,1252 @@
...
+    await stakeManager.connect(signers[5]).stake(epoch, stake);
+    stakes.push(stake);
+  });
+
+  afterEach(async () => {
+    await restoreSnapshot(snapShotId);
+    stakes = [];
+  });
+
+  it('100 epochs of constant voting and participation', async () => {
```
abhishekvispute, line 124: Are we going to run scenario tests per commit in CI, or have periodic runs ?
If we do every time CI would take significant time, 2X with gasCompare

SamAg19: As of right now, it is running every commit we make. I do agree though, CI would take a lot of time with currently how it is

abhishekvispute (approved):

hrishikeshio (approved):

PR #565, opened 2021-11-11
added revert string error in solhint

adi44:
have changed revert-string warning to error
made error messages under 32 bytes where ever required


adi44:
closing this due to redundant commits

PR #577, opened 2021-11-16
Add randao penalty to the stakechanged enum

### LINKED ISSUE #567: Add randao penalty to the stakechanged enum

GauravJain9:
fixes #567 
Added RandaoPenalty to the stakeChanged event 
In case of any randaoPenalty reason "RandaoPenalty" shall be emitted and in inactivity , 'inactivity & randao' in both these cases reason 'InactivityPenalty' must be emitted


contracts/Core/RewardManager.sol
```diff
@@ -86,6 +88,7 @@ contract RewardManager is Initializable, Constants, RewardManagerParams, IReward
         // uint256 currentStake = previousStake;
         if (newStake < previousStake) {
             stakeManager.setStakerEpochFirstStakedOrLastPenalized(epoch, stakerId);
+            if (randao) stakeManager.setStakerStake(epoch, stakerId, StakeChanged.RandaoPenalty, previousStake, newStake);
```
abhishekvispute, line 91: Hey this is not ideal 
SetStatkerStake is going to be called twice for setting same newStake

GauravJain9: Fixed @abhishekvispute

SkandaBhat: reason = randao ? StakeChanged.RandaoPenalty : StakeChanged.InactivityPenalty

stakeManager.setStakerStake(epoch, stakerId, StakeChanged.InactivityPenalty, previousStake, newStake);

GauravJain9: Hey here reason cannot be of string type as explicit conversion is needed and it shows enum cannot be converted to string.
I tried to include in enum too it's not working (type error)

GauravJain9: Hey we can't have setStakerStake outside the if block as it will be called unnecessarily even if there is no penalty and thus each time a false event will be emitted if no penalty is there.

SkandaBhat (approved):

abhishekvispute (approved):

Issue #576, opened 2021-11-16
Add an event to record sRZR transfer


adi44:
fields: stakerID, to address, amount

PR #586, opened 2021-11-24
Added event for srzr_transfer

adi44:
Following are the changes made in the codebase 
-- Added a hook in stakeManager.sol, that will emit a event when ever sRZR transfer will take place. #576 
-- Made changes in gas Bot script, now it will also compare the minimum gas consumed
-- Added a Access control on srzr_event method. Only Stoken Addresses can call them.


adi44:
The added event works as expected. It is being indexed from stakeManager contract events.

hrishikeshio (approved):

SkandaBhat (approved):

PR #591, opened 2021-11-25
AgeChanged reason added

### LINKED ISSUE #587: Do not reduce age on inactivity
Reducing it to 0 directly is harsh, have fair mech

### LINKED ISSUE #590: Add reason to AgeChange event
- [ ] Reason is required on web

### LINKED ISSUE #599: Remove redundant age change
There is issue with this age mech itself 
How there can be AgeChanged.InactivityPenalty and AgeChanged.MissedEpoch both ?

so at first in givePenalties, we call giveInactivityPenalties which says penalise this much for age 
= no of epcochs missed * 10000 

then we come to this if 
which directly reduce it to 0 

isn't this double ?
what was purpose of this if ?

_Originally posted by @abhishekvispute in https://github.com/razor-network/contracts/pull/591#discussion_r757294685_

SamAg19:
fixes #590 
fixes #599 
fixes #587

### DIFF
--- a/contracts/Core/RewardManager.sol
+++ b/contracts/Core/RewardManager.sol
@@ -91,7 +91,7 @@ contract RewardManager is Initializable, Constants, RewardManagerParams, IReward
             stakeManager.setStakerStake(epoch, stakerId, reason, previousStake, newStake);
         }
         if (newAge < previousAge) {
-            stakeManager.setStakerAge(epoch, stakerId, newAge);
+            stakeManager.setStakerAge(epoch, stakerId, newAge, AgeChanged.InactivityPenalty);
         }
     }
 
@@ -100,7 +100,7 @@ contract RewardManager is Initializable, Constants, RewardManagerParams, IReward
         Structs.Staker memory thisStaker = stakeManager.getStaker(stakerId);
         uint32 epochLastRevealed = voteManager.getEpochLastRevealed(stakerId);
         if (epochLastRevealed != 0 && epochLastRevealed < epoch - 1) {
-            stakeManager.setStakerAge(epoch, thisStaker.id, 0);
+            stakeManager.setStakerAge(epoch, thisStaker.id, 0, AgeChanged.MissedEpoch);
             return;
         }
         uint64 age = thisStaker.age + 10000;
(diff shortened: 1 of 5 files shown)


SkandaBhat (approved):

contracts/Core/RewardManager.sol
```diff
@@ -100,7 +100,7 @@ contract RewardManager is Initializable, Constants, RewardManagerParams, IReward
         Structs.Staker memory thisStaker = stakeManager.getStaker(stakerId);
         uint32 epochLastRevealed = voteManager.getEpochLastRevealed(stakerId);
         if (epochLastRevealed != 0 && epochLastRevealed < epoch - 1) {
-            stakeManager.setStakerAge(epoch, thisStaker.id, 0);
+            stakeManager.setStakerAge(epoch, thisStaker.id, 0, AgeChanged.MissedEpoch);
```
abhishekvispute, line 103: Hey I think there is issue with this age mech itself 
How there can be AgeChanged.InactivityPenalty and AgeChanged.MissedEpoch both ?

so at first in givePenalties, we call giveInactivityPenalties which says penalise this much for age 
= no of epcochs missed * 10000 

then we come to this if 
which directly reduce it to 0 

isn't this double ?
what was purpose of this if ?

abhishekvispute: @razor-network/engineering

abhishekvispute (changes requested):

contracts/Core/RewardManager.sol
abhishekvispute: looks like this if is not needed, what was its purpose 
anyways we are thinking removing this step where we reduce age directly to 0 
so should we remove this if entirely ? @hrishikeshio @SkandaBhat 

Related Issue https://github.com/razor-network/contracts/issues/587

hrishikeshio: yes but the if and return is still needed. else the following calculation will not be correct.

hrishikeshio (changes requested):

hrishikeshio (approved):

abhishekvispute (approved):

Issue #588, opened 2021-11-24
Staker can jump the queue at the time of block proposal


SamAg19:
Suppose 3 valid blocks have been proposed

```[0,1,2]```

A new block is proposed which has the wrong biggestInfluence and is placed at the following position

```[3,0,1,2]```

Another block is proposed which is a valid block but has an iteration greater than block id 2. It replaces 3 and this is the new ```sortedProposedBlocks``` array

```[4,0,1,2]```

Block id 4 jumps ahead of the rest of the blocks even though it has the highest iteration. Disincentives early block proposals as there can be a chance for a staker to jump the queue incase an invalid block is proposed

SamAg19:
Will need to have a discussion on this to find the optimal solution

PR #593, opened 2021-11-26
Test case for staker can jump the queue at the time of block proposal

shekhar2807:
#588
Staker can jump the queue at the time of block proposal if any staker proposes the block brefore him wtih wrong biggest influece id,

The following test case shows that scenario if any staker proposed with wrong biggest influence id , then in near future (epochs) that staker can disturb the queue. 

Note: This PR is supposed to fail in CI.


SkandaBhat:
Tests failing. Dismissing this request.

SkandaBhat:
Closing this as the issue has been validated.

PR #597, opened 2021-11-29
Check Epoch removed

### LINKED ISSUE #589: Remove epoch check for all except commit, reveal, propose

SamAg19:
fixes #589

### DIFF
--- a/contracts/Core/BlockManager.sol
+++ b/contracts/Core/BlockManager.sol
@@ -73,11 +73,8 @@ contract BlockManager is Initializable, BlockStorage, StateManager, BlockManager
     }
 
     //anyone can give sorted votes in batches in dispute state
-    function giveSorted(
-        uint32 epoch,
-        uint16 collectionId,
-        uint32[] memory sortedStakers
-    ) external initialized checkEpochAndState(State.Dispute, epoch, epochLength) {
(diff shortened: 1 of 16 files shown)


hrishikeshio:
Please note that the issue has tag "analysis"

Please provide reasons why its safe to remove the epoch restriction for each of these functions.

SamAg19:
Epoch restriction can be removed in withdraw function only as the razors that are going to be withdrawn have already been locked and separated from the network at the time of unstake. A withholding attack will not cause any loss to the locked funds.

Will reinstate the epoch restriction in other functions except withdraw.

hrishikeshio:
> Epoch restriction can be removed in withdraw function only as the razors that are going to be withdrawn have already been locked and separated from the network at the time of unstake. A withholding attack will not cause any loss to the locked funds.
> 
> Will reinstate the epoch restriction in other functions except withdraw.

why cant we remove epoch check in other functions?

SkandaBhat:
Delegate and unstake should not have epoch check. @SamAg19 please make the changes accordingly.

SkandaBhat (changes requested):
Please make the required changes.

hrishikeshio (approved):

abhishekvispute (approved):

PR #602, opened 2021-12-03
Staker wont be able to jump queue anymore.

### LINKED ISSUE #588: Staker can jump the queue at the time of block proposal

SamAg19:
fixes #588

### DIFF
--- a/contracts/Core/BlockManager.sol
+++ b/contracts/Core/BlockManager.sol
@@ -219,7 +219,7 @@ contract BlockManager is Initializable, BlockStorage, StateManager, BlockManager
         uint8 sortedProposedBlockslength = uint8(sortedProposedBlockIds[epoch].length);
 
         if (sortedProposedBlockslength == 0) {
-            sortedProposedBlockIds[epoch].push(0);
+            sortedProposedBlockIds[epoch].push(blockId);
             blockIndexToBeConfirmed = 0;
             return;
         }
--- a/contracts/Core/BlockManager.sol
+++ b/contracts/Core/BlockManager.sol
@@ -66,8 +66,11 @@ contract BlockManager is Initializable, BlockStorage, StateManager, BlockManager
         uint256 biggestInfluence = voteManager.getInfluenceSnapshot(epoch, biggestInfluencerId);
         if (sortedProposedBlockIds[epoch].length == 0) numProposedBlocks = 0;
         proposedBlocks[epoch][numProposedBlocks] = Structs.Block(true, proposerId, medians, iteration, biggestInfluence);
-        _insertAppropriately(epoch, numProposedBlocks, iteration, biggestInfluence);
+        bool isAdded = _insertAppropriately(epoch, numProposedBlocks, iteration, biggestInfluence);
         epochLastProposed[proposerId] = epoch;
+        if (isAdded) {
+            numProposedBlocks = numProposedBlocks + 1;
+        }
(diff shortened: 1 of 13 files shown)


contracts/Core/BlockManager.sol
```diff
@@ -138,18 +138,6 @@ contract BlockManager is Initializable, BlockStorage, StateManager, BlockManager
         _confirmBlock(epoch - 1, deactivatedCollections, stakerId);
     }
 
-    function disputeBiggestInfluenceProposed(
```
hrishikeshio, line 141: why is this removed?

hrishikeshio (changes requested):

contracts/Core/BlockManager.sol
SamAg19: a block with incorrect biggestInfluence can not be a part of the sortedProposedBlockIds. The only way i guess such a block can be a part of this is if all the block proposed have been incorrect.

hrishikeshio: thats exactly why this was added.

SamAg19: Added back https://github.com/razor-network/contracts/pull/602/commits/a1be8c7816f8525687ebcdd1da058b39c34158b9

contracts/Core/BlockManager.sol
```diff
@@ -224,14 +224,21 @@ contract BlockManager is Initializable, BlockStorage, StateManager, BlockManager
             return;
         }
 
-        for (uint8 i = 0; i < sortedProposedBlockslength; i++) {
-            // Replace : New Block has better biggest influence
-            if (proposedBlocks[epoch][sortedProposedBlockIds[epoch][i]].biggestInfluence < biggestInfluence) {
-                sortedProposedBlockIds[epoch][i] = blockId;
-                return;
+        if (proposedBlocks[epoch][sortedProposedBlockIds[epoch][0]].biggestInfluence > biggestInfluence) {
+            return;
```
hrishikeshio, line 228: at line 71, numProposed blocks is increased even though the block was not inserted.

SamAg19: Won't that be the case even if a valid block is proposed but length is greater than maxAltBlocks?

hrishikeshio: no. in that case a block is actually inserted.

SamAg19: But a block is popped off as we will have to maintain the length of maxAltBlocks and numProposedBlocks is also being used as ID for each block that has been proposed.

hrishikeshio: why increase numProposed if the block was never inserted. Its as if the block was never proposed.

SamAg19: Fixed https://github.com/razor-network/contracts/pull/602/commits/4e178a31d50dd430d69a67a545da1339b41663c7

hrishikeshio (approved):

abhishekvispute (approved):

PR #618, opened 2021-12-13
Unstake and Withdraw lock to mitigate the epsilon attack

### LINKED ISSUE #606: P+episilon attack can cause mass unstaking

SamAg19:
fixes #606


contracts/Core/StakeManager.sol
```diff
@@ -173,44 +176,61 @@ contract StakeManager is Initializable, StakeStorage, StateManager, Pause, Stake
     /// @param stakerId The Id of staker associated with sRZR which user want to unstake
     /// @param sAmount The Amount in sRZR
     function unstake(uint32 stakerId, uint256 sAmount) external initialized whenNotPaused {
+        Structs.Staker storage staker = stakers[stakerId];
+        uint32 epoch = _getEpoch(epochLength);
+        require(staker.id != 0, "staker.id = 0");
+        require(staker.stake > 0, "Nonpositive stake");
+        require(locks[msg.sender][staker.tokenAddress][LockType.Unstake].amount == 0, "Existing Unstake Lock");
+        require(locks[msg.sender][staker.tokenAddress][LockType.Withdraw].amount == 0, "Existing Withdraw Lock");
```
hrishikeshio, line 184: does it matter? I think this is not required

contracts/Core/StakeManager.sol
```diff
@@ -173,44 +176,61 @@ contract StakeManager is Initializable, StakeStorage, StateManager, Pause, Stake
...
+        require(sAmount > 0, "Non-Positive Amount");
+
+        IStakedToken sToken = IStakedToken(staker.tokenAddress);
+        require(sToken.balanceOf(msg.sender) >= sAmount, "Invalid Amount");
+
+        locks[msg.sender][staker.tokenAddress][LockType.Unstake] = Structs.Lock(sAmount, 0, epoch + lockPeriod);
```
hrishikeshio, line 190: unstakelockperiod and withdrawlockperiod should be different.

contracts/Core/StakeManager.sol
```diff
@@ -173,44 +176,61 @@ contract StakeManager is Initializable, StakeStorage, StateManager, Pause, Stake
...
         require(currentState != State.Dispute, "Unstake: NA Dispute");
         uint32 epoch = _getEpoch(epochLength);
 
         Structs.Staker storage staker = stakers[stakerId];
-        require(staker.id != 0, "staker.id = 0");
-        require(staker.stake > 0, "Nonpositive stake");
-        require(locks[msg.sender][staker.tokenAddress].amount == 0, "Existing Lock");
-        require(sAmount > 0, "Non-Positive Amount");
+        Structs.Lock storage lock = locks[msg.sender][staker.tokenAddress][LockType.Unstake];
+        require(staker.id != 0, "staker doesnt exist");
```
hrishikeshio, line 202: should be checked at top. 
https://docs.soliditylang.org/en/v0.6.11/security-considerations.html#use-the-checks-effects-interactions-pattern

contracts/Core/StakeManager.sol
```diff
@@ -173,44 +176,61 @@ contract StakeManager is Initializable, StakeStorage, StateManager, Pause, Stake
...
+        require(lock.unlockAfter != 0, "Did not unstake");
+        require(lock.unlockAfter <= epoch, "Withdraw epoch not reached");
+        require(lock.unlockAfter + withdrawInitiationPeriod >= epoch, "Initiation Period Passed"); // Can Use ExtendLock
 
         // slither-disable-next-line reentrancy-events,reentrancy-no-eth
         rewardManager.giveInactivityPenalties(epoch, stakerId);
 
         IStakedToken sToken = IStakedToken(staker.tokenAddress);
-        require(sToken.balanceOf(msg.sender) >= sAmount, "Invalid Amount");
+        require(sToken.balanceOf(msg.sender) >= lock.amount, "locked amount lost");
```
hrishikeshio, line 211: how is it possible? defeats purpose

abhishekvispute: hmm we are not taking custody of sRZRs, so one can transfer them 
mistakes purpose is only to declare that you want to withdraw this much sRZRs 
and then in withdraw we convert

contracts/Core/StakeManager.sol
```diff
@@ -173,44 +176,61 @@ contract StakeManager is Initializable, StakeStorage, StateManager, Pause, Stake
     /// @param stakerId The Id of staker associated with sRZR which user want to unstake
     /// @param sAmount The Amount in sRZR
     function unstake(uint32 stakerId, uint256 sAmount) external initialized whenNotPaused {
+        Structs.Staker storage staker = stakers[stakerId];
```
hrishikeshio, line 179: https://docs.soliditylang.org/en/v0.6.11/security-considerations.html#use-the-checks-effects-interactions-pattern

contracts/Core/StakeManager.sol
```diff
@@ -173,44 +176,61 @@ contract StakeManager is Initializable, StakeStorage, StateManager, Pause, Stake
...
+        uint32 epoch = _getEpoch(epochLength);
+        require(staker.id != 0, "staker.id = 0");
```
hrishikeshio, line 181: stakerId should be validated first. not staker.id

contracts/Core/StakeManager.sol
```diff
@@ -173,44 +176,61 @@ contract StakeManager is Initializable, StakeStorage, StateManager, Pause, Stake
...
+        require(staker.stake > 0, "Nonpositive stake");
+        require(locks[msg.sender][staker.tokenAddress][LockType.Unstake].amount == 0, "Existing Unstake Lock");
+        require(locks[msg.sender][staker.tokenAddress][LockType.Withdraw].amount == 0, "Existing Withdraw Lock");
+        require(sAmount > 0, "Non-Positive Amount");
```
hrishikeshio, line 185: input parameters should be validated first

contracts/Core/StakeManager.sol
hrishikeshio, line 190: You actually need to transfer the stoken from user. It's not really a "lock" if user just sells the srzr after locking. defeats the purpose and doesnt prevent attacks.

contracts/Core/StakeManager.sol
hrishikeshio, line 211: https://docs.soliditylang.org/en/v0.6.11/security-considerations.html#use-the-checks-effects-interactions-pattern

contracts/Core/StakeManager.sol
hrishikeshio: can still be further up. we dont need epoch before requires.

contracts/Core/StakeManager.sol
```diff
@@ -173,44 +176,65 @@ contract StakeManager is Initializable, StakeStorage, StateManager, Pause, Stake
...
+        require(sToken.balanceOf(msg.sender) >= sAmount, "Invalid Amount");
+
+        locks[msg.sender][staker.tokenAddress][LockType.Unstake] = Structs.Lock(sAmount, 0, epoch + unstakeLockPeriod);
+        emit Unstaked(msg.sender, epoch, stakerId, sAmount, staker.stake, block.timestamp);
+
+        require(sToken.transferFrom(msg.sender, address(this), sAmount), "sToken transfer failed");
+    }
+
+    function initiateWithdraw(uint32 stakerId) external initialized whenNotPaused {
         State currentState = _getState(epochLength);
```
hrishikeshio, line 197: require first.

contracts/Core/parameters/Governance.sol
```diff
@@ -64,14 +64,19 @@ contract Governance is Initializable, ACL, Constants {
         stakeManagerParams.setSlashParams(_bounty, _burn, _keep);
     }
 
-    function setWithdrawLockPeriod(uint8 _withdrawLockPeriod) external initialized onlyRole(GOVERNER_ROLE) {
+    function setUnstakeLockPeriod(uint8 _unstakeLockPeriod) external onlyRole(GOVERNER_ROLE) {
```
hrishikeshio, line 67: why was initialized remoed?

contracts/Core/StakeManager.sol
SamAg19: in this case, i think we should be checking for state first before going to other require statements. If we aren't in the right state, we should not go ahead with further checks.

contracts/Core/parameters/Governance.sol
SamAg19: by mistake. Fixed https://github.com/razor-network/contracts/pull/618/commits/48d0e7899bc1a513d218bd35afb312ae6b37a987

contracts/Core/StakeManager.sol
abhishekvispute: hmm 
though its small change
I think we should always verify input params first 
if staker id is 0 then state shouldn't be checked 
reason being we are calculating state, that's not input param 
so needlessly cal and checking that when input params are invalid, is not good approach imo

SamAg19: We have actually always checked for the state first and then for input params in all our functions. It's just that in other functions, state checks are being done with the help of a modifier and here, we aren't using that since we have got check for 2 states.

contracts/Core/StakeManager.sol
```diff
@@ -173,44 +176,65 @@ contract StakeManager is Initializable, StakeStorage, StateManager, Pause, Stake
...
+            "Initiation Period Passed"
+        ); // Can Use ExtendLock
 
         Structs.Staker storage staker = stakers[stakerId];
-        require(staker.id != 0, "staker.id = 0");
-        require(staker.stake > 0, "Nonpositive stake");
-        require(locks[msg.sender][staker.tokenAddress].amount == 0, "Existing Lock");
-        require(sAmount > 0, "Non-Positive Amount");
+        IStakedToken sToken = IStakedToken(staker.tokenAddress);
+        Structs.Lock storage lock = locks[msg.sender][staker.tokenAddress][LockType.Unstake];
```
abhishekvispute, line 212: can we move this lines variable lines up to this requires 
and in req use those variables 
that would be more efficient and clean imo
```
 require(locks[msg.sender][stakers[stakerId].tokenAddress][LockType.Unstake].unlockAfter != 0, "Did not unstake");

  ```

File: contracts/Core/StakeManager.sol [hunk 11]
```diff
@@ -173,44 +176,65 @@ contract StakeManager is Initializable, StakeStorage, StateManager, Pause, Stake
...
+        uint256 rAmount = _convertSRZRToRZR(lock.amount, staker.stake, sToken.totalSupply());
         staker.stake = staker.stake - rAmount;
 
         // Transfer commission in case of delegators
         // Check commission rate >0
         uint256 commission = 0;
         if (stakerIds[msg.sender] != stakerId && staker.commission > 0) {
             // Calculate Gain
-            uint256 initial = sToken.getRZRDeposited(msg.sender, sAmount);
+            uint256 initial = sToken.getRZRDeposited(msg.sender, lock.amount);
```
abhishekvispute, line 225: hmm this won't work as intended now 
as msg.sender don't own lock.amount now, becoz now we are transferring it in unstake to this contract.

For ex.
```
Razor depo 100
sRZR 100
Current sRZR valuation 200
```
now you are transferring it in unstake to this contract 
hence for this user 
if you unstake 100 sRZRs 
then razorDeposited for you is updated to 0 
and for this contract as 100 

https://github.com/razor-network/contracts/blob/master/contracts/tokenization/StakedToken.sol#L70

then when you calculate gain
its consider as 
current sRZR value(200) - razorDeposited(0) = 200
and commission is calculated on that 
not on 
200 - 100

abhishekvispute: I am not sure why this was not picked up in tests itself
Commission calculation will always go wrong with this change,
Are we not calculating commission in tests ?

abhishekvispute (changes requested):

contracts/Core/StakeManager.sol
SamAg19: We are calculating commission in tests. such an error never popped up

test/StakeManager.js
```diff
@@ -640,23 +704,29 @@ describe('StakeManager', function () {
...
+      const prevStake = staker.stake;
+      const totalSupply = await sToken.totalSupply();
       const rAmount = (amount.mul(staker.stake)).div(totalSupply);
+      await stakeManager.connect(signers[5]).initiateWithdraw(staker.id);
       staker = await stakeManager.getStaker(4);
 
+      const lock = await stakeManager.locks(signers[5].address, staker.tokenAddress, 1);
       const initial = await sToken.getRZRDeposited(signers[5].address, amount); // How much delegator had put for this much amount of SRZRS
       const gain = (rAmount.sub(initial)); // commission in accordance to gain
       const commission = ((gain).mul(staker.commission)).div(100);
```
abhishekvispute, line 732: hmm I think this is the reason why tests not failed with this change
we are calling contract to getRazorDeposited 
so its in line with contract but not logically

SamAg19: Yep, was just about to say the same thing.
I'll look more into this

abhishekvispute (dismissed):

abhishekvispute (dismissed):

test/StakeManager.js
hrishikeshio: resolved?

contracts/Core/StakeManager.sol
hrishikeshio: resolved?

test/StakeManager.js
SamAg19: Yep resolved

SkandaBhat:
Closing this. Please reopen at a later point. @SamAg19

abhishekvispute:
There is new attack vector possible as hit and run via withdraw due to this 
We need to see
Similar issue 
#659

contracts/Core/StakeManager.sol
```diff
@@ -286,22 +310,28 @@ contract StakeManager is Initializable, StakeStorage, StateManager, Pause, Stake
 
     /// @notice Used by anyone whose lock expired or who lost funds, and want to request withdraw
```
abhishekvispute, line 311: Please update comments on function to what the reflect now, also if possible please add on PR how this PR resolves P + e.

contracts/Core/StakeManager.sol
```diff
@@ -172,44 +175,65 @@ contract StakeManager is Initializable, StakeStorage, StateManager, Pause, Stake
...
+        uint32 epoch = _getEpoch(epochLength);
         Structs.Staker storage staker = stakers[stakerId];
-        require(staker.id != 0, "staker.id = 0");
-        require(staker.stake > 0, "Nonpositive stake");
-        require(locks[msg.sender][staker.tokenAddress].amount == 0, "Existing Lock");
-        require(sAmount > 0, "Non-Positive Amount");
+        Structs.Lock storage lock = locks[msg.sender][staker.tokenAddress][LockType.Unstake];
+        require(lock.unlockAfter != 0, "Did not unstake");
+        require(lock.unlockAfter <= epoch, "Withdraw epoch not reached");
+        require(lock.unlockAfter + withdrawInitiationPeriod >= epoch, "Initiation Period Passed"); // Can Use ExtendLock
```
abhishekvispute, line 210: hey as of there looks like three lock periods here, unstakeLockPeriod, withdrawInitiationPeriod, and withdrawLockPeriod. If I understand correctly there is unstake, then unlockAfter is = epoch + unstakeLockPeriod. then initiaite withdraw can only happen if epoch<= epoch + unstakeLockPeriod + withdrawInititationPeriod, right ? If yes please consider removing any one of it as its redundant

SamAg19: how is it redundant? withdrawInitiationPeriod is similar to what we had with withdrawReleasePeriod we have currently

contracts/lib/Structs.sol
```diff
@@ -28,9 +28,10 @@ library Structs {
     }
 
     struct Lock {
-        uint256 amount; //amount in RZR
-        uint256 commission;
-        uint256 withdrawAfter; // Can be made uint32 later if packing is possible
+        uint256 amount; //amount in sRZR/RZR
```
abhishekvispute, line 31: Yep as you mentioned I think we can do packing here

abhishekvispute: Razor supply * decimals : 10 ** 9 * 10 ** 18 = 10 ** 27
Thats still lot less than 2 ^ 256
image

abhishekvispute: I think we can have 
uint128 amount 
uint128 initial

SamAg19: in that case shouldnt we do that everywhere else? uint256 is being used predominantly for stake amounts

abhishekvispute: On commission I dont think we need to maintain so, we can leverage initial field only 
as of we have this 
```
amount
commission.  0(UnstakeLock), X(WithdrawLock)
unlockAfter
Initial              X(UnstakeLock), 0(WithdrawLock)
```
In any of locks if we have this one field is redundant 
what if we remove commssion entirely, and make it to have one lock, rather than two, hence we would save redundant ``SSTORES``

abhishekvispute: Something like this 
Unstake 
```
Create lock 
amount : sAmount
unLockAfter : unstakeLockPeriod
Initial : razorDepostied 
```
InitiateWithdraw 
check if unlockAfter is passed

```
Update lock (not creating new)
amount : rAmount
unlockAftter: withdrawLockPeriod
initial : razorDeposted (old value only) 
```

UnlockWithdraw 
```
Calculate commission here from amount - initial
and resetLock 
(In this case only one would need to be reset)
```

If we do this we will get rid of LockType and 0 values getting stored, and storing two locks onchain 
**Let me know what do you think about it** 
Sorry I didnt spot this in previous approved review

contracts/Core/StakeManager.sol
abhishekvispute: hmm maybe I am missing something, lets check this in todays call

abhishekvispute: Is it like this 
unstakeLockPeriod, : 1
withdrawInitiationPeriod, : 5
withdrawLockPeriod. : 1


epoch 0 
unstake 
unLockAfter : 1

only allow initiate withdraw when unlockAfter + 5 <= epoch 

now my question is in unstake only, couldnt we have assigned this unlockAfter + 5 value only
6 rather than 1

contracts/lib/Structs.sol
abhishekvispute: > in that case shouldnt we do that everywhere else? uint256 is being used predominantly for stake amounts

hmm on this yes we should if its allowing packing

abhishekvispute: hmm I think lets put packing on hold 
We dont want it to overflow like it did that day with influence on multiplication to find commission
Please check another suggestion of above of making one lock only and removing commission from the lock as now we have initial anyway. 
CC : @SamAg19

contracts/Core/StakeManager.sol
abhishekvispute: lets discuss this anyway on call today
as I am not sure if this is needed now

contracts/lib/Structs.sol
SamAg19: > hmm I think lets put packing on hold We dont want it to overflow like it did that day with influence on multiplication to find commission Please check another suggestion of above of making one lock only and removing commission from the lock as now we have initial anyway. CC : @SamAg19

Issue with the having one lock would be that how would extension work? how would we able to differentiate whether it is sRazor or Razor amount? both of them would have different mechanisms of giving extension penalty.

abhishekvispute: yep I think we can remove extension entirely, as now we are locking razors, 😄 
lets discuss more on call

abhishekvispute: TLDR : Removing commission from lock, and use initial

abhishekvispute (approved):

hrishikeshio (approved):

contracts/Core/StakeManager.sol
```diff
@@ -167,71 +170,90 @@ contract StakeManager is Initializable, StakeStorage, StateManager, Pause, Stake
...
+        require(sToken.transferFrom(msg.sender, address(this), sAmount), "sToken transfer failed");
+    }
+
+    /// @notice staker/delegator call initiateWithdraw() to burn their locked sRZRs and lock their RAZORS
+    // RAZORS is separated from the staker's stake but can be claimed only after withdrawLockPeriod passes
+    // after which he/she can call unlockWithdraw() to claim the locked RAZORS.
+    /// @param stakerId The Id of staker associated with sRZR which user want to initiateWithdraw
+    function initiateWithdraw(uint32 stakerId) external initialized whenNotPaused {
         State currentState = _getState(epochLength);
         require(currentState != State.Propose, "Unstake: NA Propose");
```
yohanelly95, line 204: @GauravJain9 check this

PR #620, opened 2021-12-16
can no longer delegate to staker if staker stake is less than minStake

### LINKED ISSUE #604: If staker falls below minstake then no one can delegate

SamAg19:
fixes #604

### DIFF
--- a/contracts/Core/StakeManager.sol
+++ b/contracts/Core/StakeManager.sol
@@ -147,6 +147,7 @@ contract StakeManager is Initializable, StakeStorage, StateManager, Pause, Stake
         require(stakers[stakerId].acceptDelegation, "Delegetion not accpected");
         require(_isStakerActive(stakerId, epoch), "Staker is inactive");
         require(!stakers[stakerId].isSlashed, "Staker is slashed");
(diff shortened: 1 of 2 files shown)


contracts/Core/StakeManager.sol
```diff
@@ -147,6 +147,7 @@ contract StakeManager is Initializable, StakeStorage, StateManager, Pause, Stake
         require(stakers[stakerId].acceptDelegation, "Delegetion not accpected");
         require(_isStakerActive(stakerId, epoch), "Staker is inactive");
         require(!stakers[stakerId].isSlashed, "Staker is slashed");
+        require(stakers[stakerId].stake >= minStake, "Staker stake less than minStake");
```
abhishekvispute, line 150: Hmm whats the purpose of doing this ?

abhishekvispute: Hmm got it

abhishekvispute: Do we really need this but
*Extra Check per Delegation*
This is going to be checked for each delegation 
And as a delegator I should do my own research before delegating 

Though following is rare
If stake is below minStake I can sponser staker to run
With this PR I cant do that

PR #649, opened 2022-01-17
Assign Collections Randomly :game_die:

abhishekvispute:
Epic 
Fixes #684

Child Issues 
Fixes #135 Fixes #657 Fixes #644 Fixes #643 Fixes #572 Fixes #699

## Commit
````
Commit is changed, we are going back to merkle root being committed, and are introducing random assignment
In reveal node can only reveal collections assigned to them.
Assignment is dependent on salt and secret, salt is stored in confirm block of last epoch, and can be calculated manually.
So you are right in thinking that in commit only, stakers would know their assignment
This cant be exploited as you cant know other peoples assignment given secret is "secret" :)
So cant find if you are the only one revealing some collection
Now the tree that node has to built, has to contain all the numActiveCollections (PosAwareMerkleTree Req)
But the values for the non assigned collection doesnt matter
So we suggest to node to consider votes for non-assgined to be 0 only, so they dont have to do keccakhash redundantly
Please note that this tree is PosAware, so you cant just post gigantic tree root, and pick values as per convenience in reveal

@dev Randao Removed : So there is no penalty as randao penalty.
But Random Number Service is still there, it uses salt to get random no now
Staker is only penalised for the collections they have revealed don't match with medians.
This is decided on basis of non-zero vote value.

/// Steps
/// Fetch NumActiveCollections and no of collections to assign (toAssign)
/// Calculate Salt = keccak256(abi.encodePacked(epoch, blocks[epoch].medians))
/// Find Assignments
/// Construct Tree with values for assigned and 0 for non-assigned
/// Commitment is as per
/// commitment = keccak256(abi.encodePacked(tree.root, seed))
/// where seed = keccak256(abi.encodePacked(salt, secret));

/// @dev
/// There are two cryptographic parts
/// isAssetAlloted
/// Merkle
/// Please note that if assets alloted are 2,1,1
/// a tree is constructed for 0,1,2,3,4
/// in a tree there is no repetition
/// but in a isCollectionAlloted, input array
/// you have to pass [2,1,1], here seq and repetation should be maintained

````

## Reveal 
```
In reveal, staker has to pass secret, root of tree, revealed values and seq of allocated collections
Note that for any of assigned revealed collection, the vote value cant be zero, as if would have allowed, a clever staker could have escaped penalties.
hence 
Pls note, With this PR, result value of collections, **cant be 0**.

// Input Params
// epoch, treeRevealData, secret
// treeRevealData (follwoing struct)
// struct MerkleTree {
[...]

PR #656, opened 2022-01-20
randao bug fix

### LINKED ISSUE #653: Delegator can unstake in reveal, before staker reveals, resulting in penalty and age being reset to 0 for staker
Staker commits 
Reveal Phase starts 
Before staker reveals 
Delegator withdraws

SamAg19:
fixes #653

### DIFF
--- a/contracts/Core/StakeManager.sol
+++ b/contracts/Core/StakeManager.sol
@@ -183,13 +183,16 @@ contract StakeManager is Initializable, StakeStorage, StateManager, Pause, Stake
         require(locks[msg.sender][staker.tokenAddress].amount == 0, "Existing Lock");
         require(sAmount > 0, "Non-Positive Amount");
 
-        // slither-disable-next-line reentrancy-events,reentrancy-no-eth
-        rewardManager.giveInactivityPenalties(epoch, stakerId);
-
         IStakedToken sToken = IStakedToken(staker.tokenAddress);
         require(sToken.balanceOf(msg.sender) >= sAmount, "Invalid Amount");
+        uint256 totalSupply = sToken.totalSupply();
 
-        uint256 rAmount = _convertSRZRToRZR(sAmount, staker.stake, sToken.totalSupply());
+        if ((staker.stake - _convertSRZRToRZR(sAmount, staker.stake, totalSupply)) < minStake) {
+            // slither-disable-next-line reentrancy-events,reentrancy-no-eth
+            rewardManager.giveInactivityPenalties(epoch, stakerId);
+        }
+
+        uint256 rAmount = _convertSRZRToRZR(sAmount, staker.stake, totalSupply);
         staker.stake = staker.stake - rAmount;
 
         // Transfer commission in case of delegators
--- a/contracts/Core/StakeManager.sol
+++ b/contracts/Core/StakeManager.sol
@@ -194,7 +194,7 @@ contract StakeManager is Initializable, StakeStorage, StateManager, Pause, Stake
             rewardManager.giveInactivityPenalties(epoch, stakerId);
         }
 
-        if(prevStake != staker.stake) rAmount = _convertSRZRToRZR(sAmount, staker.stake, totalSupply);
+        if (prevStake != staker.stake) rAmount = _convertSRZRToRZR(sAmount, staker.stake, totalSupply);
         staker.stake = staker.stake - rAmount;
 
         // Transfer commission in case of delegators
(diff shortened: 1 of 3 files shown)


contracts/Core/StakeManager.sol
```diff
@@ -183,13 +183,16 @@ contract StakeManager is Initializable, StakeStorage, StateManager, Pause, Stake
...
         require(sAmount > 0, "Non-Positive Amount");
 
-        // slither-disable-next-line reentrancy-events,reentrancy-no-eth
-        rewardManager.giveInactivityPenalties(epoch, stakerId);
-
         IStakedToken sToken = IStakedToken(staker.tokenAddress);
         require(sToken.balanceOf(msg.sender) >= sAmount, "Invalid Amount");
+        uint256 totalSupply = sToken.totalSupply();
+
+        if ((staker.stake - _convertSRZRToRZR(sAmount, staker.stake, totalSupply)) < minStake) {
```
SkandaBhat, line 190: `_convertSRZRToRZR(sAmount, staker.stake, totalSupply)` can instead be just `rAmount`? Code repitition

abhishekvispute (approved):
lgtm

contracts/Core/StakeManager.sol
abhishekvispute: hmm rAmount before giving penalties and after would change 
as stakers stake change
that's why we have to calculate twice I think

but it can be optimised 
have rAmount at top 
and only update when we penalise

contracts/Core/StakeManager.sol
```diff
@@ -186,13 +186,15 @@ contract StakeManager is Initializable, StakeStorage, StateManager, Pause, Stake
...
+        uint256 prevStake = staker.stake;
 
-        if ((staker.stake - _convertSRZRToRZR(sAmount, staker.stake, totalSupply)) < minStake) {
+        if ((staker.stake - rAmount) < minStake) {
             // slither-disable-next-line reentrancy-events,reentrancy-no-eth
             rewardManager.giveInactivityPenalties(epoch, stakerId);
         }
 
-        uint256 rAmount = _convertSRZRToRZR(sAmount, staker.stake, totalSupply);
+        if(prevStake != staker.stake) rAmount = _convertSRZRToRZR(sAmount, staker.stake, totalSupply);
```
abhishekvispute: hey cant this line moved after L195 only in if block ?
also we don't have to track prevStake imo
we can just make follwoing at L195, right ?

``rAmount = _convertSRZRToRZR(sAmount, staker.stake, totalSupply);``

with current impl, we will have redundant prevStake and if condition - L197 checked everytime

contracts/Core/StakeManager.sol
```diff
@@ -183,13 +183,18 @@ contract StakeManager is Initializable, StakeStorage, StateManager, Pause, Stake
...
-
         IStakedToken sToken = IStakedToken(staker.tokenAddress);
         require(sToken.balanceOf(msg.sender) >= sAmount, "Invalid Amount");
+        uint256 totalSupply = sToken.totalSupply();
+        uint256 rAmount = _convertSRZRToRZR(sAmount, staker.stake, totalSupply);
+        uint256 prevStake = staker.stake;
+
+        if ((staker.stake - rAmount) < minStake) {
+            // slither-disable-next-line reentrancy-events,reentrancy-no-eth
+            rewardManager.giveInactivityPenalties(epoch, stakerId);
```
hrishikeshio, line 194: we are supposed to do this for randao penalty and not for inactive penalty.
randao penalty is small but inactive penalty can be very big

if staker is inactive for thousands of epochs and inactivity penalty > min stake, staker can escape inactivity penalty by doing this (he will lose minstake instead)

hrishikeshio (changes requested):

contracts/Core/StakeManager.sol
SamAg19: oh ya correct. forgot about this

contracts/Core/StakeManager.sol
SamAg19: Will have to separate out randao and inactivity into 2 different functions to implement this. Randao is anyway getting removed in #649 so this can be considered as a quick fix.

What say? @abhishekvispute @SkandaBhat @hrishikeshio

abhishekvispute: hey 
cant we do this check on this line on the top of randao penalty only 
https://github.com/razor-network/contracts/blob/580af180bf20d6ca3ab487428725f15adaf9afae/contracts/Core/RewardManager.sol#L88

so inactivity will be as it is 
and for randao we will have this check

abhishekvispute: hmm but then this will be applicable for commit as well nvm

abhishekvispute: Ya then now we have two options 
1) Divide the function for only randao
2) Use above and let staker escape inactivity if > minstake
3) Disallow unstake in commit and reveal

Now if its just for quickfix : 
I think 3) is better option for a while, its not creating any new issue and a ux concern.

SamAg19: Ya makes sense. Will do this

PR #674, opened 2022-02-01
Base denominator increased to 10 million

### LINKED ISSUE #669: Increase the base deno to allow lower values for gov params
with 10000 as base, we cant go low as 0.0001
make it 10 million

SamAg19:
fixes #669


contracts/Core/RewardManager.sol
```diff
@@ -9,6 +9,7 @@ import "./interface/ICollectionManager.sol";
 import "../Initializable.sol";
 import "./storage/Constants.sol";
 import "./parameters/child/RewardManagerParams.sol";
+import "hardhat/console.sol";
```
abhishekvispute, line 12: hey it wont be included in final bytecode, as sol compiler would do tree shaking 
But even so, can we remove this 
Just for being clean sake

abhishekvispute: All else LGTM

abhishekvispute:
So just to verify
These params are updated to higher range for better precision 
Params list can be found here 
https://github.com/razor-network/contracts/blob/v0.2.0-incentnet-prod/parameters.md

penaltyNotRevealNum
SlashNums
MaxTolerance (So tolerance passed)
Base Deno

@SamAg19 please confirm

CC : @hrishikeshio @SkandaBhat

contracts/Core/RewardManager.sol
SamAg19: removed https://github.com/razor-network/contracts/pull/674/commits/a67257c9ee441c2c86295869fe5ba7093dfbba58

SkandaBhat (approved):

PR #681, opened 2022-02-07
Configure single deployment script in package.json for multiple env

### LINKED ISSUE #679: Generic Deployment script for all Environment
Right now we have different deployment script for different networks/environment in package.json ex: `npm run deploy:local`, `npm run deploy:mumbai`, and if we introduce more environment then we need to add same in package.json which is bit duplicate work.

As we are already defining those variables in `.env` file we should use from there instead for passing again as argument to deployment script. ex: `./scripts/deploy.sh mumbai` and `./scripts/deploy.sh local`

We can remove these vars from `.env` and can pass directly with npm run command ex: `npm run deploy -- ENV NETWORK`

**Additional context**
No

Shrikant1212:
Can refer Issue [679](https://github.com/razor-network/contracts/issues/679)
Fixes #679 

@SkandaBhat


SkandaBhat:
@Shrikant1212 please add the prefix "Fixes <issue number>" to the description, so that it is easier to automatically close the issue when this PR is merged.

SkandaBhat (changes requested):
please address the comment.

PR #694, opened 2022-02-10
Safe age limit and required added

### LINKED ISSUE #688: Create a safeAgeLimit and require the age isnt more than that
Currently, the governance parameter max-age does not have an upper bound, which will break maturities if the age is > 100.

Create a safeAgeLimit and require the age isnt more than that.

SamAg19:
fixes #688


abhishekvispute (approved):

contracts/Core/parameters/Governance.sol
```diff
@@ -106,6 +106,7 @@ contract Governance is Initializable, ACL, Constants {
     }
 
     function setMaxAge(uint32 _maxAge) external initialized onlyRole(GOVERNER_ROLE) {
+        require(_maxAge <= 100 * 10000, "Invalid Max Age Update");
```
SkandaBhat, line 109: Does it make sense to assign this to say `maxSafeAge`? We may end up forgetting about this require if we change the age mapping.

SkandaBhat (changes requested):

abhishekvispute (approved):

hrishikeshio (approved):

PR #698, opened 2022-02-13
[Delegator.sol ]added a new method to get result from ID and removed strict rule to create collection with same name 

### LINKED ISSUE #696: Modifications in contracts for Bridge
Delegator : 
Problem 1 :
Name => Id is mapping is stored 

Problem 2 :
Activated Collections List : There is a getter alrea

Solution :
With AAR,
We can use the IDs passed in propose.
**Expose a getter on delegator, to get the result from id.**

### LINKED ISSUE #700: Move Name to ID and setNametoID method mapping in collection manager

adi44:
fixes #696 
fixes #700


SamAg19 (approved):

contracts/Core/CollectionManager.sol
```diff
@@ -215,4 +225,9 @@ contract CollectionManager is CollectionStorage, StateManager, CollectionManager
             }
         }
     }
+
+    function _setIDName(string calldata name, uint16 _id) internal {
```
adi44, line 229: add restriction back

contracts/Core/CollectionManager.sol
```diff
@@ -176,6 +173,22 @@ contract CollectionManager is Initializable, CollectionStorage, StateManager, Co
         return collections[id];
     }
 
+    function getResult(bytes32 _name) external view override returns (uint32, int8) {
+        uint16 index = idToIndexRegistry[ids[_name]];
```
abhishekvispute, line 177: hey I think L178 to L181 and L186 to L189 is repeated 
I think what can call getResultFromID only in getResult right ?

adi44: hmm, might be possible, but in that case I would need to make getResultFromID public so we could use it here. But does it make change in gas cost consumption. what we can do make one internal function that gives result from ID and use that

abhishekvispute: hey diff between external and public on gas costs is not applicable anymore I think 
From 0.8.0
Had seen some tweets on same 
Check this
https://twitter.com/transmissions11/status/1474540805922574337?s=20&t=IiO-3kknBzYfJGn42IvDaw

abhishekvispute: So I think we can make it public without worrying about so

abhishekvispute (approved):

contracts/Core/CollectionManager.sol
adi44: sounds good

abhishekvispute (approved):

contracts/Core/CollectionManager.sol
```diff
@@ -2,15 +2,16 @@
 pragma solidity ^0.8.0;
 
 import "./interface/ICollectionManager.sol";
-import "../IDelegator.sol";
+import "./interface/IBlockManager.sol";
 import "./interface/IVoteManager.sol";
 import "./storage/CollectionStorage.sol";
+import "../Initializable.sol";
```
abhishekvispute, line 8: hey Initializable being imported twice

abhishekvispute (changes requested):

abhishekvispute (approved):

SkandaBhat (approved):

PR #712, opened 2022-02-21
Natspec added to parameters and random manager

SamAg19:


contracts/Core/RewardManager.sol
```diff
@@ -21,7 +21,8 @@ contract RewardManager is Initializable, Constants, RewardManagerParams, IReward
     IBlockManager public blockManager;
     ICollectionManager public collectionManager;
 
-    /** @param stakeManagerAddress The address of the VoteManager contract
+    /**
+     * @param stakeManagerAddress The address of the StakeManager contract
```
abhishekvispute, line 25: hey here collection manger also needs to be added as param

abhishekvispute: also please write Natspec for internal functions as well

SamAg19: I have not added natspec on internal functions as it is not going to be parsed by the solidity compiler. It is still upto us though whether we want to add natspec to internal functions

SkandaBhat: Lets add them. Will be better for readability and auditing.

contracts/Core/RewardManager.sol
```diff
@@ -81,6 +88,12 @@ contract RewardManager is Initializable, Constants, RewardManagerParams, IReward
         }
     }
 
+    /**
+     * @dev Penalties are given to stakers based their activity if they have been inactive for more than the grace period
+     * and their votes in the previous epoch compared to the medians confirmed. Penalties on botes depend upon how far were
```
abhishekvispute, line 93: I don't want to add review on English 😄 
But here it should be ``votes`` instead of ``botes``

abhishekvispute (approved):

contracts/Core/RewardManager.sol
SamAg19: fixed https://github.com/razor-network/contracts/pull/712/commits/3300506664f50da270755398db4cd653542c4b21

abhishekvispute (approved):

hrishikeshio (approved):

hrishikeshio:
what about failing gascompare? @SamAg19

PR #726, opened 2022-02-24
fix: only propose revealed assets of epoch

### LINKED ISSUE #705: Proposer shouldn't pass values for non assigned assets in this epoch
https://github.com/razor-network/contracts/pull/649#issuecomment-1042811069

### LINKED ISSUE #706: Opt: Hash - getActiveCollections inside create and update status
https://github.com/razor-network/contracts/pull/649

abhishekvispute:
Fixes #705 
Fixes #706 

Like we discussed in last arch call on AAR 
Have removed the condition that staker has to propose non revealed assets only.
Now staker only has to propose revealed assets of this epoch.
This would have broke delegator, as it was assuming everything available every epoch
So as we discussed last time, have added result mapping stored in confirm O(n)

Have renamed medianIndex to activeCollectionIndex as now median dont represent all active collections
so was misleading.


SkandaBhat:
Scenarios failing @abhishekvispute

contracts/Core/CollectionManager.sol
```diff
@@ -308,25 +296,21 @@ contract CollectionManager is Initializable, CollectionStorage, StateManager, Co
...
     /// @inheritdoc ICollectionManager
     function getIdToIndexRegistryValue(uint16 id) external view override returns (uint16) {
+        require(collections[id].active, "Query for Inactive Collection");
         return idToIndexRegistry[id];
     }
 
     /// @inheritdoc ICollectionManager
-    function getActiveCollectionsHash() external view override returns (bytes32 hash) {
-        hash = keccak256(abi.encodePacked(getActiveCollections()));
+    function getIndexToIdRegistryValue(uint16 index) external view override returns (uint16) {
```
SamAg19, line 306: Now that we are storing results for each id, why do we require the registries? Can't Index be replaced with the IDs now?

abhishekvispute: hmm nope 
merkle pos aware only works on seq digits 
which ids are not, once any is deactivated
hence reveal must happen on index 
and for same we req reg

SamAg19 (dismissed):

contracts/Core/BlockManager.sol
```diff
@@ -163,92 +146,76 @@ contract BlockManager is Initializable, BlockStorage, StateManager, BlockManager
...
-
-    function disputeForProposedCollectionIds(uint32 epoch, uint8 blockIndex) external initialized checkEpochAndState(State.Dispute, epoch) {
-        uint32 blockId = sortedProposedBlockIds[epoch][blockIndex];
+        Structs.Block memory _block = proposedBlocks[epoch][blockId];
 
-        require(proposedBlocks[epoch][blockId].valid, "Block already has been disputed");
+        // shouldnt be present
+        if (totalInfluenceRevealed == 0) {
+            bool toDispute = false;
+            for (uint256 i = 0; i < _block.ids.length; i++)
```
abhishekvispute, line 171: combine code block

contracts/Core/BlockManager.sol
```diff
@@ -163,92 +146,76 @@ contract BlockManager is Initializable, BlockStorage, StateManager, BlockManager
...
-        );
+        uint16 activeCollectionIndex = disputes[epoch][msg.sender].activeCollectionIndex;
+        // get position in block for that activeCollectionIndex
+        uint16 id = collectionManager.getIndexToIdRegistryValue(activeCollectionIndex);
+
+        Structs.Block memory _block = proposedBlocks[epoch][blockId];
+
+        uint32 proposedValue = 0;
+        for (uint256 i = 0; i < _block.ids.length; i++)
+            if (_block.ids[i] == id) {
```
abhishekvispute, line 213: ask for user to pass this index

contracts/Core/BlockManager.sol
abhishekvispute: should be absent is O(1)

abhishekvispute: Should be present can be log(n)

abhishekvispute:
```
Commit O(1)
Reveal O(n) n= toAssign
Propose O(n) n=maxAltBlocks
Dispute
	1. Value
		GiveSorted O(n) n= no of unique votes for this Coll
		FinDispute O(1)  ask user to pass position
	2. BiggestStakerId passed O(1)
	3. Ids present O(log n)/absent O(1),  n= no of coll revealed this epoch
	4. Ids sorted O(1)
	
Confirm 
	O(n) n= no of collection revealed this epoch


```

SamAg19 (dismissed):

SkandaBhat (dismissed):
lgtm

SamAg19 (approved):

SamAg19 (approved):

hrishikeshio (approved):

PR #729, opened 2022-02-24
docs: natspec added to delegator, block manager and vote manager

SamAg19:


contracts/Core/BlockManager.sol
```diff
@@ -38,21 +65,26 @@ contract BlockManager is Initializable, BlockStorage, StateManager, BlockManager
...
-
-    /// @dev The IDs being passed here, are only used for disputeForNonAssignedCollection
-    /// for delegator, we have seprate registry
-    /// If user passes invalid ids, disputeForProposedCollectionIds can happen
-
+    /**
+     * @notice elected proposer proposes block.
+     * we use a probabilistic method to elect stakers weighted by stake
+     * protocol works like this.
+     * select a staker pseudorandomly (not weighted by anything)
```
abhishekvispute, line 72:
```
 select a staker pseudorandomly (not weighted by anything)
 that staker then tosses a biased coin
 ```
hmm is this true anymore ? its old comment
do we select staker before tossing coin ?
we toss a coin and select proposer

also this line
```
note that only one staker or no stakers selected in each iteration.
```

File: contracts/Core/VoteManager.sol [hunk 2]
```diff
@@ -33,6 +57,20 @@ contract VoteManager is Initializable, VoteStorage, StateManager, VoteManagerPar
         collectionManager = ICollectionManager(collectionManagerAddress);
     }
 
+    /**
+     * @notice stakers query the jobs in each collection, aggregate and instead of revealing them instantly,
```
abhishekvispute, line 61: hmm this is not true
you don't have to fetch for each collection 
Only for assigned ones 

also next line tho its corrected line after
they don't submit hash of their results 
```
hash of their results which becomes their commitment 
```
The correct flow is you fetch numActiveCollections, find assignments, fetch those results, construct Merkle Pos aware tree, commit hash of root and seed

contracts/Core/VoteManager.sol
```diff
@@ -56,6 +94,13 @@ contract VoteManager is Initializable, VoteStorage, StateManager, VoteManagerPar
         }
     }
 
+    /**
+     * @notice staker reveal the votes that they had committed to the protocol in the commit state.
+     * Stakers would only reveal the collections they have been allocated, the rest of their votes wont matter
```
abhishekvispute, line 99:
```
Stakers would only reveal the collections they have been allocated, the rest of their votes won't matter
```
This line is ambiguous.

Please mention that order of votes passed matter
And you will be only able to reveal what's assigned to you 
Values for others don't matter is correct, but only in tree, you cant pass them as votes

Also is possible 
Specify what staker pass in that Merkle tree struct

contracts/Core/interface/IVoteManager.sol
```diff
@@ -4,31 +4,85 @@ pragma solidity ^0.8.0;
...
+     * @param epoch in which the staker reveal this value
+     * @param stakerId id of the staker
+     * @param medianIndex index of the collection
+     * @return vote value
+     */
     function getVoteValue(
         uint32 epoch,
         uint32 stakerId,
-        uint16 assetId
+        uint16 medianIndex
```
abhishekvispute, line 30: please replace this with activeCollectionIndex
Its not medianIndex any more, as block is not given to have everything

contracts/Core/VoteManager.sol
SamAg19: Will remove the word `each` here but the thought process here was to write a generic statement of what happens in commit and in dev, mention the technical parts of the committing

contracts/Core/interface/IVoteManager.sol
SamAg19: cant do that unless your PR is merged. will get linting error if i make this change right now

contracts/Core/BlockManager.sol
abhishekvispute: cool we can remove so, it was never their

abhishekvispute (approved):

SkandaBhat (approved):

PR #746, opened 2022-03-11
[Skale Changes] Changed Block.number to Block.timestamp

adi44:
For skale, changed block.number to block.timestamp. State and epoch calculation is dependent on block.timestamp now.


contracts/Core/StakeManager.sol
```diff
@@ -296,9 +297,10 @@ contract StakeManager is Initializable, StakeStorage, StateManager, Pause, Stake
     function unstake(uint32 stakerId, uint256 sAmount) external initialized whenNotPaused {
         require(sAmount > 0, "Non-Positive Amount");
         require(stakerId != 0, "staker.id = 0");
+        // slither-disable-next-line timestamp
```
abhishekvispute, line 300: hmm how stakers stake is related to timestamp 
slither is suggesting issue here as well ?

abhishekvispute (approved):

contracts/Core/StakeManager.sol
adi44: yeah it was suggesting here also

adi44: we can go through slither warnings of this PR tomorrow while we discuss other slither warnings

abhishekvispute: hmm cool

SamAg19 (approved):

PR #843, opened 2022-06-16
fix: transfer razor only during testnet deployment

### LINKED ISSUE #818: [Post Deployment Script] Only transfer tokens in testnets: make the network check explicit
In https://github.com/razor-network/contracts/blob/master/migrations/src/postDeploymentSetup.js#L36, we need to be explicit in mentioning the network as "testnet" instead of !== "mainnet".

This needs to be done in v1 as well.

rajkharvar:
fixes #818

### DIFF
--- a/migrations/src/postDeploymentSetup.js
+++ b/migrations/src/postDeploymentSetup.js
@@ -76,7 +76,7 @@ module.exports = async () => {
   const stakerAddressList = STAKER_ADDRESSES.split(',');
 
   // Only transfer tokens in testnets
-  if (NETWORK !== 'mainnet') {
+  if (NETWORK === 'testnet') {
     // Add new instance of StakeManager contract & Deployer address as Minter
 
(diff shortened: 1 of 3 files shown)


migrations/src/postDeploymentSetup.js
```diff
@@ -76,7 +76,7 @@ module.exports = async () => {
   const stakerAddressList = STAKER_ADDRESSES.split(',');
 
   // Only transfer tokens in testnets
-  if (NETWORK !== 'mainnet') {
+  if (NETWORK === 'testnet') {
```
yohanelly95, line 79: Do we need to define the testnet networks explicitly here?

SkandaBhat: Introduce a new env variable, NETWORK_TYPE. Use that here. It will be either "MAINNET" or "TESTNET".

yohanelly95 (approved):
+1

PR #931, opened 2022-08-09
feat: razor token contract modification

### LINKED ISSUE #930: Modify RAZOR token contract
Changes to the razor token contract would need to be done when deploying to the skale for mainnet.

https://docs.skale.network/ima/1.3.x/managing-erc20#_1_reviewmodify_the_token_contract
To do the following modifications as mentioned here by skale, the following issues are required to be addressed:

1. We may need remove the RAZOR token contract in this repo and make the changes in this repo https://github.com/razor-network/token-contract due to solidity version discrepancy.
2. The openzeppelin version being used skale in the example is v2.5 whereas our contracts use v4.5. Because of this, we wont be able to make use of `ERC20Mintable` as mentioned in the example due to it being deprecated post v2.5. To be compatible with v4.5, we would need to use https://github.com/OpenZeppelin/openzeppelin-contracts/blob/v4.5.0/contracts/token/ERC20/presets/ERC20PresetMinterPauser.sol
3. Changes to the post deployment script may need to be done depending upon what changes are implemented based on the above 2 points

SamAg19:
fixes #930


SamAg19:
Creating a separate repo for this

PR #938, opened 2022-08-19
fix: mainnet config added and overrides removed

SamAg19:


SkandaBhat (approved):

hrishikeshio:
tests are failing. Find a way to solve issue without breaking CI.

PR #943, opened 2022-09-04
feat: storing votes and voteWeights with collectionid instead of leafid

### LINKED ISSUE #940: Use collectionID instead of leafID in all vote mappings after verifying from merkle tree
for context , check #861

SamAg19:
fixes #940


contracts/Core/BlockManager.sol
```diff
@@ -194,23 +194,23 @@ contract BlockManager is Initializable, BlockStorage, StateManager, BlockManager
...
      */
     function giveSorted(
         uint32 epoch,
-        uint16 leafId,
+        uint16 collectionId,
         uint256[] memory sortedValues
     ) external initialized checkEpochAndState(State.Dispute, epoch, buffer) {
-        require(leafId <= (collectionManager.getNumActiveCollections() - 1), "Invalid leafId");
-        uint256 medianWeight = voteManager.getTotalInfluenceRevealed(epoch, leafId) / 2;
+        require(collectionId <= (collectionManager.getNumActiveCollections()), "Invalid collectionId");
```
rajkharvar, line 205: @SamAg19 Suppose if the collectionIds are 1,2,3,4. Only collectionId 3 and 4 are active then `numActiveCollections` will be 2. Now when we try to pass collectionId as 3 which is actually correct then the contract will revert with the message "Invalid collectionId".

Can we change `collectionId <= (collectionManager.getNumActiveCollections())` to `collectionManager.getCollectionStatus(collectionId)` ?

SkandaBhat: @SamAg19 this needs to be fixed.

SamAg19: acknowledged
Fixed https://github.com/razor-network/oracle-contracts/pull/943/commits/0e97483a74c37258410e446c16d96fd2d658ecb6

rajkharvar (approved):

PR #996, opened 2024-02-08
feat: timestamp added and updated for each collection post block confirmation

### LINKED ISSUE #989: Add timestamp to each collection
## Problem Statement

The current implementation does not store the timestamp at which the latest result of a particular collection is set. Clients require a timestamp to authenticate whether the result being reported is the latest or not. 

## Solution
Add `timestamp` to each collection and update it when new values are reported in each collection.

SamAg19:
## Problem Statement

The current implementation does not store the timestamp at which the latest result of a particular collection is set. Clients require a timestamp to authenticate whether the result being reported is the latest or not. 

fixes #989 

## Solution

- When the block is confirmed, the mapping of the collection id to timestamp `latestResultTimestamp` is updated along with the result of the collection. 
- When we try to get the result through the collection manager from the `getResult` function, it will return power, result and timestamp

### DIFF
--- a/contracts/Core/BlockManager.sol
+++ b/contracts/Core/BlockManager.sol
@@ -473,8 +473,8 @@ contract BlockManager is Initializable, BlockStorage, StateManager, BlockManager
     }
 
     /// @inheritdoc IBlockManager
-    function getLatestResults(uint16 id) external view override returns (uint256) {
-        return latestResults[id];
+    function getLatestResults(uint16 id) external view override returns (uint256, uint256) {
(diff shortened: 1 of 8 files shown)


SkandaBhat:
Please add valid description in both PR and issue, and re-open.

SamAg19:
Updated description for the PR and issue @SkandaBhat

yohanelly95 (approved):

rajkharvar (approved):

PR #1002, opened 2024-08-19
feat: timestamp added and updated for each collection post block confirmation

### LINKED ISSUE #989: Add timestamp to each collection

SamAg19:
## Problem Statement

The current implementation does not store the timestamp at which the latest result of a particular collection is set. Clients require a timestamp to authenticate whether the reported result is the latest. 

fixes #989 

## Solution

- When the block is confirmed, the mapping of the collection id to timestamp `latestResultTimestamp` is updated along with the result of the collection. 
- When we try to get the result through the collection manager from the `getResult` function, it will return power, result and timestamp

### DIFF
--- a/contracts/Core/interface/ICollectionManager.sol
+++ b/contracts/Core/interface/ICollectionManager.sol
@@ -70,17 +70,19 @@ interface ICollectionManager {
      * @notice returns the result of the collection based on the name sent by the client
      * @param _name the name of the collection in bytes32
      * @return result of the collection
+     * @return result timestamp of the collection
      * @return power of the resultant collection
      */
-    function getResult(bytes32 _name) external view returns (uint256, int8);
+    function getResult(bytes32 _name) external view returns (uint256, int8, uint256);
 
     /**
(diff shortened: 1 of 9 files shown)


contracts/Core/interface/ICollectionManager.sol
```diff
@@ -70,17 +70,19 @@ interface ICollectionManager {
      * @notice returns the result of the collection based on the name sent by the client
      * @param _name the name of the collection in bytes32
      * @return result of the collection
+     * @return result timestamp of the collection
```
rajkharvar, line 73: @SamAg19 can you update this as timestamp of the collection result last updated

yohanelly95 (approved):

rajkharvar (approved):

hrishikeshio (approved):

PR #1020, opened 2025-02-19
oracle contracts fix ci

yohanelly95:


rajkharvar (dismissed):

contracts/Core/StakeManager.sol
```diff
@@ -684,6 +684,7 @@ contract StakeManager is Initializable, StakeStorage, StateManager, Pause, Stake
     }
 
     /**
+     * @param stakerId Id of the staker for whom you want to get the maturity
```
yohanelly95: @SamAg19 should I remove this change ? only stick to CI related fixes

SamAg19: ya better to remove it

SamAg19 (changes requested):

ashish10677 (dismissed):

rajkharvar (approved):

